PR #11, opened 2021-10-05
feat: navigation bar 

### LINKED ISSUE #8: [FEATURE]: Navigation bar
## Summary

- Creation of the responsive navigation bar that is always visible and changes self information accordingly to the "page" that is shown to the user. It should have the project name, the page name or board title (editable) and an area with an avatar and the username that triggers a dropdown menu.

### Basic example

In this section it's visible an example of the navbar at the dashboard page

image

### Motivation

To be able to navigate between pages and be able to know some information about the current page or current board

nunocaseiro:
Relates to #8 

image

## Proposed Changes

  - Adding a navigation bar that shows the project name, the title of page , the username and an avatar that triggers a dropdown menu.

This changes are proposed to close #8

### DIFF
--- a/frontend/components/Layout/Layout.tsx
+++ b/frontend/components/Layout/Layout.tsx
@@ -1,10 +1,12 @@
 import { styled } from "../../stitches.config";
+import NavBar from "../NavBar/NavBar";
 
 const Main = styled("main", { display: "block", px: "$40" });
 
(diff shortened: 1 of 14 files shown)


danielgek (approved):
Humm not sure about the title context is the best way to go about it but i think it can go in as is and if needed we change it later ;)
Awesome work :tada:

PR #23, opened 2021-10-08
feat: create board dialog

### LINKED ISSUE #21: [FEATURE]: create board action
## Summary

This issue is intended to create a feature that allows the user to create a new board and insert the name of the board. To do this a button will be created in dashboard that triggers a dialog to enter the name of the board 

## User story it relates to

#16 

### Basic example

image

### Motivation

We need this feature to allow the user create a new board

nunocaseiro:
Relates to #21

## Screenshots

Screenshot 2021-10-08 at 15 11 35

## Proposed Changes

  - A button in dashboard that triggers a modal - radix dialog
  - A modal with text field and a save button to create a new board

This changes are proposed to close #21

### DIFF
--- a/frontend/store/store.tsx
+++ b/frontend/store/store.tsx
@@ -1,5 +1,6 @@
 import React, { Dispatch, useContext } from "react";
-import useStateMachine, { Nullable, State, Event } from "./useStoreStateMachine";
+import { Nullable } from "../utils/types";
+import useStateMachine, { State, Event } from "./useStoreStateMachine";
 
 type ContextType = {
   state: State<string, string>;


frontend/components/NavBar/Title.tsx
```diff
@@ -1,15 +1,21 @@
...
+import { useStoreContext } from "../../store/store";
 import Text from "../Primitives/Text";
 
 const Title: React.FC = () => {
-  const titleCtx = useContext(TitleContext);
+  const router = useRouter();
+  const {
+    state: { title },
+  } = useStoreContext();
+  const isBoardPage = router.pathname.startsWith("/boards/");
```
pffigueiredo, line 10: It would probably be helpfull if we extracted page names into a separate file and import it from there

frontend/components/NavBar/Title.tsx
```diff
@@ -1,15 +1,21 @@
...
 
-  return (
+  const pageTitle = (
     <Text fontWeight="semiBold" size="xl">
-      {titleCtx.title}
+      {title}
     </Text>
   );
+
+  return <>{!isBoardPage && pageTitle}</>;
```
pffigueiredo, line 18: I can see you wanted to do an abstraction here, but I would rather not to have this kind of conditionals, as it is a simple case and I can see it being overcomplicated in a near futures, with a lot of if's.

frontend/components/Primitives/Box.tsx
```diff
@@ -0,0 +1,17 @@
+import { styled } from "../../stitches.config";
+import Flex from "./Flex";
+
+const Box = styled("div", Flex, {
+  boxShadow: "0 2px 10px rgb(62 62 82 / 30%)",
+  variants: {
+    clickable: {
+      true: {
+        "&:hover": {
```
pffigueiredo, line 9: Use the `:hover` media query, to check if the device has hover capacities, else it will show hover effects in phone mobiles, which is always kinda ugly.

You can see it being defined [here](https://github.com/radix-ui/design-system/blob/master/stitches.config.ts)

frontend/components/Primitives/Dialog/DialogContent.tsx
```diff
@@ -0,0 +1,58 @@
...
+const StyledContent = styled(Content, Flex, {
+  backgroundColor: "white",
+  borderRadius: "$6",
+  boxShadow: "hsl(206 22% 7% / 35%) 0px 10px 38px -10px, hsl(206 22% 7% / 20%) 0px 10px 20px -15px",
+  position: "fixed",
+  top: "50%",
+  left: "50%",
+  transform: "translate(-50%, -50%)",
+  width: "90vw",
+  maxWidth: "450px",
```
pffigueiredo, line 22: Not sure what you want to do here, but you will probably get away in a more declarative way with:
`width: max(90vw, 450px);`, which will pick the bigger value (if I understood what you wanted correctly)

frontend/components/Primitives/IconButton.tsx
```diff
@@ -0,0 +1,95 @@
+import { styled } from "../../stitches.config";
+
+const IconButton = styled("button", {
```
pffigueiredo, line 3: I wonder if we wanna have a BaseButton and use that as the base for both this and the `Button` component

frontend/pages/dashboard.tsx
```diff
@@ -1,14 +1,18 @@
...
-    titleCtx.setTitle("Dashboard");
-  }, [titleCtx]);
+    dispatch({ type: "setTitle", val: "Dashboard" });
+  }, [dispatch]);
 
-  return <>Dashboard</>;
+  return (
+    <>
+      <CreateBoard />
+    </>
```
pffigueiredo, line 15: There is no need for the `<></>` here

frontend/store/store.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useReducer, useContext } from "react";
+
+type Action = { type: "setTitle"; val: string };
+type Dispatch = (action: Action) => void;
+type State = StateType;
+
+type StateType = {
+  title: string;
+};
```
pffigueiredo, line 9: No need to have both `StateType` and `State`

frontend/store/store.tsx
```diff
@@ -0,0 +1,45 @@
...
+    default: {
+      throw new Error(`Unhandled action type: ${action.type}`);
+    }
+  }
+};
+
+const Context = React.createContext<ContextType | undefined>(undefined);
+
+const StoreProvider: React.FC = ({ children }) => {
+  const [state, dispatch] = useReducer(storeReducer, { title: "" });
```
pffigueiredo, line 27: There is a better way IMO to type reducers, here is a sample I made, take a look ;) 

https://gist.github.com/pffigueiredo/d92091b444e92d828f3a30af58c211d0

frontend/components/Dashboard/CreateBoardModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+);
+
+const CreateBoardModal: React.FC = () => {
+  return (
+    <Dialog>
+      {Trigger}
+      {Content}
+    </Dialog>
+  );
+};
```
jomifepe, line 44: Should we be instantiating these as function components? Or is there a reason to have them be rendered at compile time? 😛

```suggestion
const Trigger = () => (
  <DialogTrigger clickable align="center" direction="column">
    <Text size="xl">Add retro board</Text>
    <PlusIcon />
  </DialogTrigger>
);

const FooterContainer = styled("div", Flex);

const Content = () => (
  <DialogContent dialogTitle="New board" direction="column" justify="center">
    <TextField placeholder="Board name" css={{ fontSize: "$xl", width: "50%" }} />
    <FooterContainer justify="center">
      <Button size="1" variant="blue" css={{ width: "20%", mt: "$16" }}>
        Save
      </Button>
    </FooterContainer>
  </DialogContent>
);

const CreateBoardModal: React.FC = () => {
  return (
    <Dialog>
      <Trigger />
      <Content />
    </Dialog>
  );
};
```

jomifepe:
Great job, keep it up 💪

frontend/utils/PagesNames.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+const GetPageName = (url: string): string => {
+  return names[url];
+};
+
+export const CheckIsPageName = (url: string): boolean => {
+  return url.startsWith("/boards");
+};
+
+export default GetPageName;
```
pffigueiredo, line 15: Don't export default this, rule of thumb, if in doubt always do a named export `export const...`

Only export default when you are exporting the name of the file, so it's clear and expected enough

frontend/utils/PagesNames.tsx
```diff
@@ -0,0 +1,15 @@
+interface NamesType {
+  [key: string]: string;
+}
+
+const names: NamesType = { "/": "Dashboard" };
```
pffigueiredo, line 5: This should be backwards, the key should be the page name, so it's easier to consume.

frontend/store/useStoreStateMachine.tsx
```diff
@@ -0,0 +1,34 @@
+import { Dispatch, Reducer, useReducer } from "react";
+
+export type Nullable<T> = T | null | undefined;
```
pffigueiredo, line 3: Don't export this `type` from this file. This can be used throught the whole application and as such, should be inside a  `tsUtils` or whatever general file

danielgek (approved):
LGTM :fire: but also let's wait for @pffigueiredo re review ;)

Awesome work guys !!

pffigueiredo (approved):
Looking good, let's merge this fellow 💪

PR #40, opened 2021-11-03
feat: register user form - fe

### LINKED ISSUE #34: [FEATURE]: Register user
## Summary

- A form for users to sign up 

## User story it relates to

#15 

### Basic example

image

### Motivation

We need this feature to allow the users fill in with their data and then sign up

nunocaseiro:
Relates to #34 

## Screenshots 

Screenshot 2021-11-03 at 20 32 54

Screenshot 2021-11-03 at 20 43 24

## Proposed Changes

  - Installation of Next 12.0.2
  - Enabled the rust compliler
  - Auth page to login and register. 
  - Register form using radix-tabs. 
  - Input validations with yup and react-hook-form.
  - Hover card with radix-hover
  - Animation onfocus input
  - Replace fetch with axios as it works better with react-query to handle errors in the response provided by the server

 This pull request is intended to close #34


frontend/components/auth/CompoundFieldSet.tsx
```diff
@@ -0,0 +1,113 @@
...
+const StyledEyeClosedIcon = styled(EyeNoneIcon, { size: "$20" });
+const ListItem = styled("li", Text, { fontSize: "$8" });
+
+const CompoundFieldSet: React.FC<{
+  label: string;
+  id: string;
+  inputType: string;
+  errors: FieldError | undefined;
+  register: UseFormRegister<User>;
+}> = ({ label, id, inputType, errors, register }) => {
```
pffigueiredo, line 30: Create an interface for this component's props, so that it is more readable and easier to maintain

frontend/components/auth/CompoundFieldSet.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const [showPw, setShowPw] = useState(inputType !== "password");
+
+  const hoverCard = id === "password" && errors && (
```
pffigueiredo, line 33: This booleans logic should live in the place where you are actually rendering this. 

If someone uses a component called `hoverCard` they won't expect it to have any logic attached, it should be only JSX. The part of show/hide, should be down there ;)

frontend/components/auth/RegisterForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+        .matches(
+          /^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#$%^&*])(?=.{7,})/,
+          "Weak password, please check the info card."
+        ),
+      passwordConf: yup
+        .string()
+        .oneOf([yup.ref("password"), null], "Passwords must match.")
+        .required("Please enter a valid password."),
+    })
+    .required();
```
pffigueiredo, line 40: I believe you can move this object's schema out of the component, so that it doesn't get re-defined in each render

frontend/components/auth/CompoundFieldSet.tsx
```diff
@@ -0,0 +1,113 @@
...
+const StyledEyeOpenIcon = styled(EyeOpenIcon, { size: "$20" });
+const StyledEyeClosedIcon = styled(EyeNoneIcon, { size: "$20" });
+const ListItem = styled("li", Text, { fontSize: "$8" });
+
+const CompoundFieldSet: React.FC<{
+  label: string;
+  id: string;
+  inputType: string;
+  errors: FieldError | undefined;
+  register: UseFormRegister<User>;
```
pffigueiredo, line 29: You can use the native `formProvider` that `react-hook-form` exposes to fetch these errors and register properties, instead of passing them around ;) 

Take a look [here](https://react-hook-form.com/api/useformcontext)

nunocaseiro: top! thank you

pffigueiredo (commented):
LGTM 🔥

pffigueiredo (approved):

PR #50, opened 2021-11-16
feature: create, get and update board

### LINKED ISSUE #47: [FEATURE]: Create and get boards
## Summary

- Rest api endpoints that allows to create, retrieve and update the retro boards 

## User story it relates to

#16 
#17 

### Motivation

We need this to show the boards list to the users in frontend dashboard. To create a new board and also visualize the content of each board.

nunocaseiro:
Relates to #47

## Proposed Changes

  - Board controller with multiples endpoints such as create board, get boards, get board by id and update locked board
  - Board, column and card entities
  - Board, column and card dtos to also validate the incoming data
 
This pull request is intended to close #47


backend/src/boards/entity/board.entity.ts
```diff
@@ -0,0 +1,18 @@
+import { Exclude } from 'class-transformer';
+import { Column, Entity } from 'typeorm';
+import ColumnEntity from './column.entity';
+import { Content } from './content';
+
+@Entity('boards')
+class BoardEntity extends Content {
+  @Column({ nullable: false })
+  createdBy: string;
```
rpvsilva, line 9: Since you are doing a relation for the user here, you could use the `typeorm` [relations](https://typeorm.io/#/relations) feature

backend/src/boards/entity/card.entity.ts
```diff
@@ -0,0 +1,12 @@
+import { Column, Entity } from 'typeorm';
+import { Content } from './content';
+
+@Entity('cards')
+class CardEntity extends Content {
+  @Column({ nullable: false })
+  createdBy: string;
```
rpvsilva, line 7: Same as above

backend/src/boards/boards.service.ts
```diff
@@ -0,0 +1,83 @@
...
+    private boardsRepository: Repository<BoardEntity>,
+  ) {}
+
+  async create(boardData: BoardDto) {
+    if (boardData.password) {
+      boardData.password = await encrypt(boardData.password);
+    }
+    const newBoard = this.boardsRepository.create(boardData);
+    await this.boardsRepository.save(newBoard);
+    return newBoard;
```
rpvsilva, line 23:
```suggestion
    return this.boardsRepository.save(newBoard);
```

backend/src/boards/boards.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  async getBoardFromRepo(boardId: string) {
+    const board = await this.boardsRepository.findOne({
+      where: {
+        _id: new ObjectId(boardId),
+      },
+    });
+    if (!board) {
+      throw new HttpException(BOARD_NOT_FOUND, HttpStatus.NOT_FOUND);
```
rpvsilva, line 44: Line line !33 you are returning an exceptions and in !44 you are throwing one, just use the throw for exceptions and do it line 
```javascript
if(boards) return boards;

throw ...
```

rpvsilva (changes requested):

backend/src/boards/boards.controller.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { BoardsController } from './boards.controller';
+
+describe('BoardsController', () => {
+  let controller: BoardsController;
+
+  beforeEach(async () => {
+    const module: TestingModule = await Test.createTestingModule({
+      controllers: [BoardsController],
+    }).compile();
```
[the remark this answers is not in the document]
nunocaseiro: sure! it was auto generated by nest cli :p

backend/src/boards/entity/board.entity.ts
nunocaseiro: I think we need to change from typeorm to mongoose because the first is more focused on SQL databases, dont support relations with mongo. The nestjs documentation also suggest using mongoose with mongodb as the most popular MongoDB object modeling tool. 
source: [https://docs.nestjs.com/techniques/mongodb#mongo](url)

nunocaseiro: Typeorm relations don't work with mongodb as this is a nosql db but I can change the string type to user type. However we don't obtain nothing more from that :)

backend/src/boards/boards.service.ts
```diff
@@ -0,0 +1,75 @@
...
+      },
+    });
+    if (boards) return boards;
+    throw new HttpException(BOARDS_NOT_FOUND, HttpStatus.NOT_FOUND);
+  }
+
+  async getBoardFromRepo(boardId: string) {
+    const board = await this.boardsRepository.findOne({
+      where: {
+        _id: new ObjectId(boardId),
```
asantos00, line 38: pardon my random jump into this PR, but this service shouldn't be coupled with MongoDB at all, it's the repository's responsibility to interact with MongoDB and thus know that ObjectId needs to be instantiated

backend/src/boards/boards.service.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (board) return board;
+    throw new HttpException(BOARD_NOT_FOUND, HttpStatus.NOT_FOUND);
+  }
+
+  async getBoard(id: string, password?: string) {
+    const board = await this.getBoardFromRepo(id);
+    if (!board.locked) return board;
+
+    const isPasswordMatched = await compare(password, board.password);
+    if (board.locked && password && isPasswordMatched) return board;
```
asantos00, line 50: this board.locked is redundant, if the code gets here, it will already by locked because of the condition on line 47

backend/src/boards/dto/column.dto.ts
```diff
@@ -0,0 +1,25 @@
+import { IsString, IsNotEmpty, ValidateNested } from 'class-validator';
+import { v4 as uuidv4 } from 'uuid';
+import { Transform, TransformFnParams, Type } from 'class-transformer';
+import CardDto from './card.dto';
+
+export class ColumnDto {
+  @IsString()
+  _id: string = uuidv4();
```
asantos00, line 8: Isn't this automatically created by mongo? This is completely out of curiosity

backend/src/boards/entity/content.ts
```diff
@@ -0,0 +1,10 @@
+import { Column, ObjectIdColumn } from 'typeorm';
+import { ObjectId } from 'mongodb';
+
+export abstract class Content {
+  @ObjectIdColumn()
+  _id: ObjectId;
```
asantos00, line 6: Just to clarify, is this _id the same thing in which we used the `uuidv4` function call above?

backend/src/constants/jwt.ts
```diff
@@ -0,0 +1,7 @@
+export const JWT_REFRESH_TOKEN_SECRET = 'JWT_REFRESH_TOKEN_SECRET';
```
asantos00, line 1: Not that it's super relevant, but ES6 Symbols can be used here. They generate unique keys and make it pretty much impossible to use the string version, because it doesn't exist: https://riptutorial.com/javascript/example/15827/implementing-enums-using-symbols

Just a matter of curiosity

backend/src/boards/dto/column.dto.ts
nunocaseiro: Yes, it is automatically created by mongo when the main document is created. I just post a body with board data with columns and cards as an array so they are nested objects. As I need the Id in the frontend of this nested documents, it's necessary to generate it manually. The ids are not mandatory but I am using the ids of nested objects as key prop on map method of render in frontend. [Here](https://github.com/chnirt/chnirt-demo-nest-typeorm-mongodb/blob/master/src/modules/user/user.entity.ts), [here](https://medium.com/@chnirt/how-do-i-practice-with-nestjs-nestjs-typeorm-mongodb-9e407818a296) and [here](https://stackoverflow.com/questions/10144852/how-can-i-create-unique-ids-for-embedded-documents-in-mongodb) they are also using uuid or objectId to generate id before insert :) Maybe I'm doing something wrong

backend/src/boards/entity/content.ts
nunocaseiro: Here I am defining the columns of an entity

backend/src/boards/dto/column.dto.ts
asantos00: Wait, so the main reason we're creating these ids is to have a key prop on the front-end? 
This manual generation of ids, together with other problems you are having (namely not being able to establish proper relations on the ORM) should trigger a discussion on if we're using the correct data storage (SQL vs NOSQL).

However, even if we decide to go with generating id's manually (besides all the flags it might raise), it shouldn't be done this way. What if you change your MongoDB configuration and the id's are not uuid v4 anymore? Will you have to change all your code? Doesn't make much sense, right?

Manual generation of ids should be done like the following:

```js
const mongodb = require("mongodb")

const newId = mongodb.ObjectId().toString()
```

backend/src/boards/entity/content.ts
asantos00: So, why is it a string above and here an objectId?

backend/src/boards/dto/column.dto.ts
nunocaseiro: Yes, atm the created ids are just for the key prop in frontend. I agree with you, we need to discuss if we should continue with nosql. In the meanwhile I will replace uuid v4 with your suggestion. Thank you for the help

asantos00: It seems strange to me that, to fix a front-end problem, to be more specifically a react problem, we're adding a field on the database. These things shouldn't be this coupled, it also doesn't seem like the right solution. 
There are many ways to generate unique keys on the front-ent that don't depend on the database, don't we agree?

nunocaseiro: Yes, I agree with you, my bad. im still with the relational db mindset and I didn't know what the best approach would be. We don't need an extra field in db. So, if i need to generate a unique id in frontend do you have any suggestion? Is uuid v4 a good ideia?

asantos00: You can probably generate the keys on the frontend by adding multiple fields, like title + color + index, even hashing them if needed. This is the better solution. Then, if not possible, you can use multiple packages, which uuid is one of them, to generate keys

nunocaseiro: Thank you! :)

backend/src/boards/boards.controller.ts
```diff
@@ -0,0 +1,60 @@
...
+  async createBoard(@Body() boardData: BoardDto) {
+    const user = await this.usersService.getByEmail(boardData.createdBy.email);
+    if (user) {
+      return this.boardService.create(boardData);
+    }
+  }
+
+  @UseGuards(JwtAuthenticationGuard)
+  @Get()
+  async boards(@Req() request: RequestWithUser) {
```
rpvsilva, line 36: Since you removed the `await` you can remove the `async`. Do it everywhere

backend/src/boards/boards.service.ts
```diff
@@ -0,0 +1,70 @@
...
+        password = null;
+      }
+      const hashedPw = password ? await encrypt(password) : null;
+      await this.boardsRepository.update(boardId, {
+        locked,
+        password: hashedPw,
+      });
+      return await this.getBoardFromRepo(boardId);
+    }
+    throw new HttpException(BOARD_NOT_FOUND, HttpStatus.UNAUTHORIZED);
```
rpvsilva, line 68: Just make the validation to throw the exceptions first and do the logic after it like:

```javascript
if (board.createdBy.email !== email) throw new HttpException(BOARD_NOT_FOUND, HttpStatus.UNAUTHORIZED);
```

rpvsilva (changes requested):

backend/src/boards/boards.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    throw new HttpException(describe(BOARD_NOT_FOUND), HttpStatus.NOT_FOUND);
+  }
+
+  async getBoard(id: string, password?: string) {
+    const board = await this.getBoardFromRepo(id);
+    if (!board.locked) return board;
+
+    const isPasswordMatched = await compare(password, board.password);
+    if (password && isPasswordMatched) return board;
+
```
rpvsilva, line 50: Noticed now that after I requested that changes here you are checking by the password argument after using it to compare... Sorry about that

backend/src/boards/boards.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (locked === false) {
+      password = null;
+    }
+    const hashedPw = password ? await encrypt(password) : null;
+    await this.boardsRepository.update(boardId, {
+      locked,
+      password: hashedPw,
+    });
+    return await this.getBoardFromRepo(boardId);
```
rpvsilva, line 75: Remove the await 😂 Sorry for being so annoying

nunocaseiro: its ok bro! thanks for the help. sometimes im blinded ahah

rpvsilva (approved):

rpvsilva:
Nice work here 🔥

PR #53, opened 2021-11-23
refactor: airbnb eslint rules

nunocaseiro:
## Proposed Changes

  - Added esling airbnb rules and adapted the code


.gitignore
```diff
@@ -1,5 +1,9 @@
...
 package-lock.json
 yarn.lock
+.DS_Store
+
 .env
-.DS_Store
\ No newline at end of file
+.env.local
+.env.development
+.env.production
```
rpvsilva: Since you removed all `.env` files, can you add a `.env.example` where you put the keys used as an example to new developers that join this project?

backend/src/constants/config.ts
```diff
@@ -1 +1,3 @@
-export const PORT = process.env.BACKEND_PORT;
+const PORT = process.env.BACKEND_PORT;
+
+export default PORT;
```
rpvsilva, line 3: Since this is a config file why are you defaulting the `PORT`?

backend/src/constants/encrypt.ts
```diff
@@ -1 +1,3 @@
-export const SALT = 10;
+const SALT = 10;
+
+export default SALT;
```
rpvsilva, line 3: Same as above

rpvsilva (changes requested):

backend/src/constants/config.ts
nunocaseiro: Yeah, I understand what you say. The eslint rule "prefer default export" is warning me. Should i disable it in this file?

backend/src/constants/config.ts
```diff
@@ -1 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
```
rpvsilva, line 1: Instead of disabling the rule, you could remove that rule, I think the developer could choose if wants to use an `export default` or not. Just my opinion 😂

nunocaseiro: Yes, i agree with you. I just disabled it

rpvsilva (approved):

PR #54, opened 2021-11-24
docs: readme - how to use. Env example files

nunocaseiro:
## Proposed Changes

  - Added "How to run" section in readme
  - Added  an `.env.example` file into each application folder


gfdias (approved):
👍

Issue #57, opened 2021-11-25
[FEATURE]: Websocket client


nunocaseiro:
## Summary

Implementation the connection to the websocket server and develop a way to send and receive messages.

## User story it relates to

#55

### Motivation

We need this to implement real time interactions when the users are seeing the same board page.

PR #58, opened 2021-11-30
feature: websocket gateway

nunocaseiro:
Relates to #57

## Proposed Changes

  - Websocket gateway with socket.io
  - Method to handle the socket connections, mapping the websocket client id to the board room
  - When the user moves some card or column, the frontend will send a message via websocket that is handled by the developed method in backend. This receives a payload of changes, transform the board with them and updates the board in database. Finally propagates the changes to all users that are seeing the same board.


rpvsilva (changes requested):

backend/src/boards/boards.controller.ts
```diff
@@ -49,7 +49,19 @@ export default class BoardsController {
 
   @Post(':id')
   getBoard(@Req() request, @Body() passwordDto?: BoardPasswordDto) {
-    return this.boardService.getBoard(request.params.id, passwordDto.password);
+    return this.boardService.getBoardWithPassword(
+      request.params.id,
+      passwordDto.password,
```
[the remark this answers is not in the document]
nunocaseiro: I didn't have the tsconfig rule enabled so the answer is no. I enabled it and handled the errors. Thanks for the feedback

backend/src/main.ts
```diff
@@ -6,6 +6,6 @@ import { PORT } from './constants/config';
 async function bootstrap() {
   const app = await NestFactory.create(AppModule, { cors: true });
   app.useGlobalPipes(new ValidationPipe({ transform: true }));
-  await app.listen(PORT);
+  if (PORT) await app.listen(PORT);
```
rpvsilva, line 9: Instead of verify if PORT is defined could you just use a default port? this way we know that the backend runs every time, and do a `console.log` telling that the app is running at X port: `API is running at http://localhost:${PORT}`

backend/src/main.ts
```diff
@@ -1,11 +1,13 @@
...
 import AppModule from './app.module';
 import { PORT } from './constants/config';
 
 async function bootstrap() {
+  const logger: Logger = new Logger('AppGateway');
   const app = await NestFactory.create(AppModule, { cors: true });
   app.useGlobalPipes(new ValidationPipe({ transform: true }));
-  await app.listen(PORT);
+  await app.listen(PORT ?? 3200);
+  logger.log(`Application is running at http://localhost:${PORT}`);
```
rpvsilva, line 11: If you don't set a PORT it will log as undefined right? Check for the PORT before and set it to a variable and then use it here

nunocaseiro: I changed the config file. Btw, the application never starts with `PORT` undefined because before it starts, it also check if the .env file has the required env vars through the joi schema validation that is in `app.module.ts` file :)

rpvsilva (approved):
LGTM 🔥  Nice job

PR #60, opened 2021-12-07
feature: real time interactions

### LINKED ISSUE #57: [FEATURE]: Websocket client

nunocaseiro:
Relates to #57 

## Demo

https://user-images.githubusercontent.com/90208434/145241654-fe75c29b-fd7d-4df5-8e70-0210f3909680.mp4

## Proposed Changes

  - Use `socket.io` to connect with websocket server and send the changes made on specific retro board and propagate them to all users that are seeing it at the same time.
  - Refactor storeStateMachine to handle the actual board state and transformations 
  - Transform board file 
  - Next-auth package updated and refactor the jwt sign secret
  
This pull request is intendend to close #57


frontend/components/Board/Column.tsx
```diff
@@ -59,17 +56,26 @@ const Column: React.FC<ColumnBoardType> = ({ title, cards, column, index, column
...
           </Circle>
-          <Droppable droppableId={column._id ?? UNDEFINED} type="card">
-            {(prov) => (
-              <CardsList direction="column" ref={prov.innerRef} {...prov.droppableProps}>
-                <InnerList columns={columns} cards={cards} color={column.color} />
-                {prov.placeholder}
-              </CardsList>
-            )}
-          </Droppable>
+          {column._id && (
```
pffigueiredo, line 62: Isn't this always true already?

frontend/hooks/useBoard.tsx
```diff
@@ -1,10 +1,31 @@
...
+import { BoardType } from "../types/board";
+import { getBoard, getBoards, getBoardWithAuth, postBoard } from "../api/boardService";
+import UseBoardType from "../types/useBoard";
 
-const useBoard = (): UseBoardType => {
+const useBoard = (
+  autoFetchBoard: boolean,
+  autoFetchBoards: boolean,
+  boardId?: string
+): UseBoardType => {
```
pffigueiredo, line 12: Change this signature to something like `boardId, {authFetchBoard, autoFetchBoards}`, the booleans are clearly side options and should have their names attached when passing in an argument for better understand what happening when reading through the code.

Also, this should have some default values from my point of view

frontend/hooks/useBoard.tsx
```diff
@@ -1,10 +1,31 @@
...
+  const { data: session } = useSession({ required: false });
+
+  const fetchBoard = useQuery(
+    ["board", { id: boardId }],
+    () =>
+      session && boardId
+        ? getBoardWithAuth({ id: boardId, token: session.accessToken })
+        : getBoard(boardId),
+    { enabled: autoFetchBoard, refetchOnWindowFocus: autoFetchBoard }
+  );
```
pffigueiredo, line 23: What happens if there is no boardId?

frontend/types/boardChanges.ts
```diff
@@ -0,0 +1,15 @@
+export interface CardChanges {
+  colIdToRemove: string;
+  colIdToAdd: string;
+  cardIdxToRemove: number;
+  cardIdxToAdd: number;
+  type: "CARD" | string;
```
pffigueiredo, line 6: Isn't there a better type than `| string` here?

frontend/types/boardChanges.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+export interface ColumnChanges {
+  colIdxToRemove: number;
+  colIdxToAdd: number;
+  type: "COLUMN" | string;
```
pffigueiredo, line 11: Isn't there a better type than `| string` here?

pffigueiredo (changes requested):

frontend/types/boardChanges.ts
nunocaseiro: I can omit `| string`. I forgot to do that

frontend/hooks/useBoard.tsx
nunocaseiro: Yeah, I got your point. I will refactor that

frontend/hooks/useBoard.tsx
nunocaseiro: Thanks for the suggestion. I will follow that!

frontend/components/Board/Column.tsx
nunocaseiro: yes, but droppableId  is throwing an error saying that cannot be undefined...

backend/src/helper/transformBoard.ts
```diff
@@ -0,0 +1,55 @@
...
+  board: BoardT,
+  changes: CardChangesT,
+) => {
+  const { colIdToAdd, colIdToRemove, cardIdxToAdd, cardIdxToRemove } = changes;
+  const colToRemove = board.columns.find((col) => col._id === colIdToRemove);
+  const colToAdd = board.columns.find((col) => col._id === colIdToAdd);
+  const cardToAdd = colToRemove?.cardsOrder[cardIdxToRemove];
+  if (cardToAdd && colToAdd) {
+    colToRemove?.cardsOrder.splice(cardIdxToRemove, 1);
+    colToAdd?.cardsOrder.splice(cardIdxToAdd, 0, cardToAdd);
```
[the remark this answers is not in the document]
nunocaseiro: Thanks! Im going to create the suggested util methods. I thought I can mutate the array without problems since I am cloning the board to a new one one. Am I wrong?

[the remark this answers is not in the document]
pffigueiredo: From what I see in this function, it looks like you take a `board` parameter and mutate it, but if you are cloning it, that's good ;)

PR #62, opened 2021-12-10
feat: update title board and delete board

### LINKED ISSUE #61: [FEATURE]: Update board title and delete board
## Summary

- Implementation of the api route in the board controller and the correspondent method in board service to allow the edition of each board title. 
- Implementation of the api route in the board controller and the correspondent method in board service to allow the deletion of one board

## User story it relates to

#19 
#20

### Motivation

We are doing this to allow the user to edit the board title and delete a board

nunocaseiro:
Relates to #61

## Proposed Changes

  - Two api routes: the first to update the board title and the second to delete the board
  - Service methods for updating the board title and deleting the board from db
  - The above methods check if the user is the owner of the board and if true the respective behavior is applied 
 
This pull request is intendend to close #61


backend/src/boards/boards.controller.ts
```diff
@@ -77,4 +78,24 @@ export default class BoardsController {
       request.user.email,
     );
   }
+
+  @UseGuards(JwtAuthenticationGuard)
+  @Delete(':id')
+  async deleteBoard(@Req() request: RequestWithUser) {
```
rpvsilva, line 84: Don't need to be asynchronous

backend/src/boards/boards.controller.ts
```diff
@@ -77,4 +78,24 @@ export default class BoardsController {
...
+    return this.boardService.deleteBoard(request.params.id, request.user.email);
+  }
+
+  @UseGuards(JwtAuthenticationGuard)
+  @Patch(':id/updateTitle')
+  async updateTitle(
```
rpvsilva, line 90: Same as above

backend/src/boards/boards.service.ts
```diff
@@ -105,4 +117,18 @@ export default class BoardsService {
     const result = await this.boardsRepository.update(payload.id, newBoard);
     return result.raw.result.ok === 1;
   }
+
+  async deleteBoard(id: string, email: string) {
+    const board = await this.getBoardFromRepo(id);
+    if (board.createdBy.email === email) {
+      const result = await this.boardsRepository.delete(id);
+      if (result.raw.result.ok === 1) {
```
rpvsilva, line 125:
```suggestion
      const { raw: { result } } = await this.boardsRepository.delete(id);
      if (result.ok === 1) {
```

frontend/pages/boards/[boardId].tsx
```diff
@@ -124,8 +125,8 @@ const Board: React.FC<{ boardId: string }> = ({ boardId }) => {
         <Droppable droppableId="all-columns" direction="horizontal" type="COLUMN">
           {(provided) => (
             <Container {...provided.droppableProps} ref={provided.innerRef}>
-              {board.columnsOrder.map((columnId, index) => {
-                const column = board.columns.find((col) => col._id === columnId);
+              {board.columnsOrder.map((columnId: string, index: number) => {
+                const column = board.columns.find((col: ColumnType) => col._id === columnId);
```
Comment (rpvsilva) [inline frontend/pages/boards/[boardId].tsx:129]: I think the typings are not needed here. You need them if the `columnsOrder` is not well typed

backend/src/boards/boards.service.ts
```diff
@@ -49,9 +51,19 @@ export default class BoardsService {
...
-    return this.boardsRepository.save(boardData);
+  async updateTitle(id: string, email: string, title: string) {
+    const board = await this.getBoardFromRepo(id);
+    if (email !== board.createdBy.email)
+      throw new HttpException(
+        describeExceptions(EMAIL_DONT_MATCH),
+        HttpStatus.UNAUTHORIZED,
+      );
+    const result = await this.boardsRepository.update(id, { ...board, title });
+    if (result.raw.result.ok === 1) return board;
```
rpvsilva, line 62:
```suggestion
    const { raw: { result } } = await this.boardsRepository.update(id, { ...board, title });
    if (result.ok === 1) return board;
```

backend/src/boards/boards.service.ts
```diff
@@ -49,9 +51,19 @@ export default class BoardsService {
...
-  updateBoard(boardData: BoardEntity) {
-    delete boardData._id;
-    return this.boardsRepository.save(boardData);
+  async updateTitle(id: string, email: string, title: string) {
+    const board = await this.getBoardFromRepo(id);
+    if (email !== board.createdBy.email)
+      throw new HttpException(
+        describeExceptions(EMAIL_DONT_MATCH),
+        HttpStatus.UNAUTHORIZED,
+      );
```
rpvsilva, line 60: Can you extract this logic into a method since you are making the same code more than 1 time

rpvsilva (changes requested):

rpvsilva (approved):

PR #76, opened 2022-01-26
feat: add, edit, delete cards

### LINKED ISSUE #75: [FEATURE]: add, edit, delete cards
## Summary

- Implementation of add, edit and delete card features

## User story it relates to

#72 

### Motivation

We need this to allow the creation, updating and deletion of cards by the user

nunocaseiro:
Relates to #75

## Proposed Changes

  - HOC to protect the routes that require autentication
  - Refactoring the code in order to use the board's new model 
  - Add card feature
  - Edit card feature
  - Remove card feature
  - Replace yup by zod
  - Alert dialog for confirming the delete action
  - Transformboard helper to modify optimistically the board when a card is moved and dropped
  - Added the request to add, edit, delete the cards in the backend
  - Added the request to update the card position in the backend
  - Refactor nextauth to save the user's id
  - Package.json updated
  
This pull request is intendend to close #75


backend/src/models/boards/boards.service.ts
```diff
@@ -231,7 +230,10 @@ export default class BoardsService {
           },
         },
         {
-          arrayFilters: [{ 'c._id': cardId }, { 'item._id': cardItemId }],
+          arrayFilters: [
+            { 'c._id': cardId },
```
RubenMCCarreira, line 234: what is `c._id` ? can you be more explicit pls?

frontend/types/next-auth.d.ts
```diff
@@ -7,32 +7,33 @@ declare module "next-auth" {
     user: {
       name: string;
       email: string;
+      id: string;
```
RubenMCCarreira, line 10: in this new type fies, I saw `_id` and now I see just `id`. these types have to be different, or you could keep the consistency?

backend/src/models/boards/boards.service.ts
nunocaseiro: image

As we can see in the above picture, I am updating the text of a card [line 228] but I need to filter the array to find the card that I want. So I used  a letter "c" to identify a element of the card's array. What I am doing in array filters is saying that in the array of cards of that column I want the card "c" with that Id.

frontend/utils/fetchData.tsx
```diff
@@ -8,13 +9,29 @@ export const instance = axios.create({
   },
 });
 
+const nonNeededToken = ["/auth/login", "/auth/refresh"];
+
+instance.interceptors.request.use(async (config) => {
+  const { url, headers } = config;
+  if (url && headers) {
+    if (!nonNeededToken.includes(url)) {
```
RubenMCCarreira, line 17: can you joint the two statements ?

frontend/types/next-auth.d.ts
nunocaseiro: The other file is in backend. This is frontend. I really don't like `_id` but in mongo is the default :)

frontend/utils/fetchData.tsx
nunocaseiro: yes, ofc

backend/src/models/boards/boards.service.ts
RubenMCCarreira: oh ok ok, I understood. but using just `c`, to a new guy like me in the project, I have no ideia what is it xD

nunocaseiro: Sure, I will rename it to `card` instead of a letter xd

RubenMCCarreira (dismissed):

frontend/types/next-auth.d.ts
RubenMCCarreira: oh ok, sry xD

backend/src/models/boards/boards.service.ts
RubenMCCarreira: ty

RubenMCCarreira (approved):

PR #87, opened 2022-02-24
feat: upvote and downvote

### LINKED ISSUE #86: [FEATURE]: Upvote and downvote cards
## Summary

- Implementation of the api routes and respective methods to upvote and downvote

## User story it relates to

#85

nunocaseiro:
Relates to #86

## Proposed Changes

- Create and delete vote services
- Votes controller with post and delete methods
- Create and delete applications (use-cases)

This pull request is intended to close #86

### DIFF
--- a/backend/src/libs/constants/database.ts
+++ b/backend/src/libs/constants/database.ts
@@ -1,1 +1,1 @@
-export const DATABASE_URI = 'database.uri';
+export const DATABASE_URI_KEY = 'database.uri';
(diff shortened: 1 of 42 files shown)


backend/src/libs/constants/database.ts
```diff
@@ -0,0 +1 @@
+export const DATABASE_URI = 'database.uri';
```
rpvsilva, line 1: You could change this to `DATABASE_URI_KEY` because it's not the real URI

rpvsilva (approved):
LGTM 🔥  Great job 😉

PR #125, opened 2022-04-08
feat: dashboard statistics and get all boards with pagination

### LINKED ISSUE #123: [FEATURE]: Dashboard statistics and boards pagination
## Summary

- Method that count the users, teams and boards of the authenticated user
- Method that retrieve a list of boards depending of the user role  with pagination

## User story it relates to

#118 

### Motivation

We need this to show the user's boards in the dashboard and boards page

nunocaseiro:
Relates to #123

## Proposed Changes

  - Method in auth controller to retrieve some statistics about the authenticated user
  - Get all boards service method depending of the user's role and atual page in the frontend. 
     - If dashboard -> all boards of user of the last 3 months
     - if boards page and the user is super admin -> all boards 
     - if boards page and the user is not super admin -> all boards of user
 - Get all boards with pagination
 
This pull request is intended to close #123

### DIFF
--- a/backend/src/infrastructure/config/jwt.register.ts
+++ b/backend/src/infrastructure/config/jwt.register.ts
@@ -8,7 +8,7 @@ import {
 export const JwtRegister = JwtModule.registerAsync({
   imports: [ConfigModule],
   inject: [ConfigService],
-  useFactory: async (configService: ConfigService) => ({
+  useFactory: (configService: ConfigService) => ({
     secret: configService.get(JWT_ACCESS_TOKEN_SECRET),
     signOptions: {
       expiresIn: `${configService.get(JWT_ACCESS_TOKEN_EXPIRATION_TIME)}s`,
--- a/backend/src/infrastructure/config/jwt.register.ts
+++ b/backend/src/infrastructure/config/jwt.register.ts
@@ -1,0 +1,17 @@
+import { ConfigModule, ConfigService } from '@nestjs/config';
+import { JwtModule } from '@nestjs/jwt';
+import {
+  JWT_ACCESS_TOKEN_SECRET,
+  JWT_ACCESS_TOKEN_EXPIRATION_TIME,
+} from '../../libs/constants/jwt';
+
+export const JwtRegister = JwtModule.registerAsync({
+  imports: [ConfigModule],
+  inject: [ConfigService],
+  useFactory: async (configService: ConfigService) => ({
+    secret: configService.get(JWT_ACCESS_TOKEN_SECRET),
+    signOptions: {
+      expiresIn: `${configService.get(JWT_ACCESS_TOKEN_EXPIRATION_TIME)}s`,
+    },
+  }),
+});
--- a/backend/src/modules/auth/controller/auth.controller.ts
+++ b/backend/src/modules/auth/controller/auth.controller.ts
@@ -93,9 +93,11 @@ export default class AuthController {
   @Get('/dashboardStatistics')
   async getDashboardHeaderInfo(@Req() request: RequestWithUser) {
     const { _id: userId } = request.user;
-    const usersCount = await this.getUserApp.countUsers();
-    const teamsCount = await this.getTeamsApp.countTeams(userId);
-    const boardsCount = await this.getBoardApp.countBoards(userId);
+    const [usersCount, teamsCount, boardsCount] = await Promise.all([
+      this.getUserApp.countUsers(),
+      this.getTeamsApp.countTeams(userId),
+      this.getBoardApp.countBoards(userId),
+    ]);
     return { usersCount, teamsCount, boardsCount };
   }
 }
--- a/backend/src/modules/auth/controller/auth.controller.ts
+++ b/backend/src/modules/auth/controller/auth.controller.ts
@@ -11,7 +11,6 @@ import {
   BadRequestException,
   Param,
 } from '@nestjs/common';
-import { EmailParam } from '../../../libs/dto/param/email.param';
 import LocalAuthGuard from '../../../libs/guards/localAuth.guard';
 import RequestWithUser from '../../../libs/interfaces/requestWithUser.interface';
 import JwtRefreshGuard from '../../../libs/guards/jwtRefreshAuth.guard';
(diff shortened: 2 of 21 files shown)


backend/src/infrastructure/config/jwt.register.ts
```diff
@@ -0,0 +1,17 @@
...
+import { JwtModule } from '@nestjs/jwt';
+import {
+  JWT_ACCESS_TOKEN_SECRET,
+  JWT_ACCESS_TOKEN_EXPIRATION_TIME,
+} from '../../libs/constants/jwt';
+
+export const JwtRegister = JwtModule.registerAsync({
+  imports: [ConfigModule],
+  inject: [ConfigService],
+  useFactory: async (configService: ConfigService) => ({
```
rpvsilva, line 11:
```suggestion
  useFactory: (configService: ConfigService) => ({
```

backend/src/modules/auth/controller/auth.controller.ts
```diff
@@ -84,4 +93,14 @@ export default class AuthController {
...
     return !!found;
   }
+
+  @UseGuards(JwtAuthenticationGuard)
+  @Get('/dashboardStatistics')
+  async getDashboardHeaderInfo(@Req() request: RequestWithUser) {
+    const { _id: userId } = request.user;
+    const usersCount = await this.getUserApp.countUsers();
+    const teamsCount = await this.getTeamsApp.countTeams(userId);
+    const boardsCount = await this.getBoardApp.countBoards(userId);
```
rpvsilva, line 103:
```suggestion
    const [usersCount, teamsCount, boardsCount] = await Promise.all([
      this.getUserApp.countUsers(),
      this.getTeamsApp.countTeams(userId),
      this.getBoardApp.countBoards(userId),
    ]);
```

backend/src/modules/boards/controller/boards.controller.ts
```diff
@@ -54,10 +56,25 @@ export default class BoardsController {
...
+
+  @UseGuards(JwtAuthenticationGuard)
+  @Get()
+  getAllBoards(
+    @Req() request: RequestWithUser,
+    @Query() { page, size }: PaginationParams,
+  ) {
+    const { _id: userId, isSAdmin } = request.user;
+    if (isSAdmin)
+      return this.getBoardApp.getBoards('allBoards', userId, page, size);
```
rpvsilva, line 76:
```suggestion
    if (isSAdmin) {
      return this.getBoardApp.getBoards('allBoards', userId, page, size);
    }
```

backend/src/modules/boards/controller/boards.controller.ts
```diff
@@ -54,10 +56,25 @@ export default class BoardsController {
...
-  @Get()
-  boards(@Req() request: RequestWithUser) {
+  @Get('/dashboard')
+  getDashboardBoards(
+    @Req() request: RequestWithUser,
+    @Query() { page, size }: PaginationParams,
+  ) {
     const { _id: userId } = request.user;
-    return this.getBoardApp.getAllBoards(userId);
+    return this.getBoardApp.getBoards('dashboard', userId, page, size);
```
rpvsilva, line 65: You don't need to destructure the user object

backend/src/modules/boards/services/get.board.service.ts
```diff
@@ -1,32 +1,163 @@
...
+      .find({ user: userId })
+      .select('board')
+      .distinct('board')
       .lean()
       .exec();
   }
 
+  async getAllBoardIdsAndTeamIdsOfUser(userId: string) {
+    const boardIds = await this.getAllBoardsIdsOfUser(userId);
+    const teamIds = await this.getTeamService.getTeamsOfUser(userId);
```
rpvsilva, line 31:
```suggestion
    const [boardIds, teamIds] = await Promise.all([
      this.getAllBoardsIdsOfUser(userId),
      this.getTeamService.getTeamsOfUser(userId)
    ])
```

backend/src/modules/boards/services/get.board.service.ts
```diff
@@ -1,32 +1,163 @@
...
+    userId: string,
+  ) {
+    const { boardIds, teamIds } = await this.getAllBoardIdsAndTeamIdsOfUser(
+      userId,
+    );
+
+    if (option === 'dashboard') {
+      const last3Months = new Date(
+        new Date().setMonth(new Date().getMonth() - 3),
+      );
```
rpvsilva, line 46:
```suggestion
      const now = new Date();
      const last3Months = new Date(now.setMonth(now.getMonth() - 3));
```

backend/src/modules/boards/services/get.board.service.ts
```diff
@@ -1,32 +1,163 @@
...
+    return {
+      $and: [
+        { isSubBoard: false },
+        { $or: [{ _id: { $in: boardIds } }, { team: { $ne: null } }] },
+      ],
+    };
+  }
+
+  async getBoards(
+    option: 'dashboard' | 'allBoards' | 'myBoards',
```
rpvsilva, line 72: I think this logic shouldn't be in server side, you just need to be concerned about the limit and offset sent by the client, the client is the one that needs to be concerned about what he wants to get

rpvsilva (changes requested):

backend/src/modules/boards/services/get.board.service.ts
```diff
@@ -1,32 +1,171 @@
...
+        select: 'user role -board',
+        populate: {
+          path: 'user',
+          select: 'firstName lastName',
+        },
+      })
+      .lean({ virtuals: true })
+      .exec();
+
+    return { boards, hasNextPage, page };
```
rpvsilva, line 132: here you are doing 2 requests each time to check if there is a next page, let's merge it and get a better solution for it

rpvsilva (approved):

backend/src/modules/boards/services/get.board.service.ts
nunocaseiro: yes, I think we can do it with the aggregator method from mongo. We need to put more effort on it. Let's merge it then

rpvsilva (approved):

PR #127, opened 2022-04-11
chore: improve some code

rpvsilva:
## Proposed Changes

  - Improved some logic
  - Instead of using an object on params just called ex: @Param('socketId')
  - Made some spacing to be more readable
  - Removed some unnecessary variables

### DIFF
--- a/backend/src/modules/boards/controller/boards.controller.ts
+++ b/backend/src/modules/boards/controller/boards.controller.ts
@@ -16,29 +16,28 @@ import BoardDto from '../dto/board.dto';
 import JwtAuthenticationGuard from '../../../libs/guards/jwtAuth.guard';
 import RequestWithUser from '../../../libs/interfaces/requestWithUser.interface';
 import { TYPES } from '../interfaces/types';
-import { CreateBoardApplication } from '../interfaces/applications/create.board.application.interface';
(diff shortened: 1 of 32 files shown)


backend/src/modules/boards/controller/boards.controller.ts
```diff
@@ -47,59 +46,57 @@ export default class BoardsController {
...
   @UseGuards(JwtAuthenticationGuard)
   @Get(':boardId')
-  async getBoard(@Param() params: BaseParam, @Req() request: RequestWithUser) {
-    const {
-      user: { _id: userId },
-    } = request;
-    const { boardId } = params;
-    const board = await this.getBoardApp.getBoard(boardId, userId);
+  async getBoard(
+    @Param('boardId') boardId: string,
```
nunocaseiro, line 64: I had this params classes to validate the data that is sent by the endpoint like we do in dtos. Don't you think we could validate if the sent "boardId" is a mongoId with class-validator? This allow us to avoid some exceptions when we do some operations in mongo. If a string is passed with more or less than 12 bytes and it hasn't a mongoDB format an error is thrown. This is also applied to other changes in the multiples controllers

nunocaseiro (changes requested):

nunocaseiro (approved):
LGTM :rocket:

Issue #137, opened 2022-04-23
[FEATURE]: Create split retro board


nunocaseiro:
## Summary

As a user I want to create a main board and split my team users into multiple sub teams. I also want to be able to customize some board configurations

### Basic example

![image](https://user-images.githubusercontent.com/90208434/164943133-6bcf53f3-9fb0-4fca-b9b2-ef242e486b40.png)

PR #140, opened 2022-04-28
feat: generate main and sub boards

nunocaseiro:
Relates to #137

## Screenshots 

image

## Proposed Changes

  - Generate main board
  - Generate sub boards from the xgeeks team
  - Quick edit settings (changing sub teams names or max number of people in each team)
  - Create board request to the backend
  - Feature to replace the responsible of each team


frontend/hooks/useCreateBoard.tsx
```diff
@@ -0,0 +1,155 @@
...
+      }
+
+      for (let i = 0; i < maxTeams; i++) {
+        const newBoard = generateSubBoard(i + 1);
+        splitedUsers[i][Math.floor(Math.random() * splitedUsers[i].length)].role =
+          BoardUserRoles.RESPONSIBLE;
+        newBoard.users = splitedUsers[i];
+        subBoards.push(newBoard);
+      }
+      return subBoards;
```
RubenMCCarreira, line 91:
```suggestion
const subBoards: BoardToAdd[] = [];
const splitedUsers: BoardUserToAdd[][] = [];

let availableUsers;

new Array (teamMembers.length).forEach((_, i) => {
  const teamUser = getRandomUser(availableUsers);
  if (splitedUsers[i]?.length === maxUsers) break;

  if (i >= maxTeams) i = 0;
  splitedUsers.push({
    user: teamUser.user,
    role: BoardUserRoles.MEMBER,
    votesCount: 0,
  })

  availableUsers = teamMembers.filter((user) => user.user._id !== teamUser.user._id);
})

new Array(maxTeams).forEach((_, i) => {
  const newBoard = generateSubBoard(i + 1);
  splitedUsers[i][Math.floor(Math.random() * splitedUsers[i].length)].role =
    BoardUserRoles.RESPONSIBLE;
  newBoard.users = splitedUsers[i];
  subBoards.push(newBoard);
})

return subBoards;
```

RubenMCCarreira: try this code pls, and check if it does the expected

nunocaseiro: Thanks for the tip but the break can't be used in a for each

RubenMCCarreira: replace for a `return;`

nunocaseiro: there is no way to break a for each

nunocaseiro: The first loop was wrong implemented. I noticed the `let j` was missing. However I tried with your suggestion `new Array` but I can't change de index in a for each so I cannot achieve the same behaviour. Is this regular for loop a wrong approach? I can try to find a better way to do this

nunocaseiro: I simplified the for loop and i also used your suggestion for the last loop to generate sub boards

RubenMCCarreira (approved):

PR #169, opened 2022-05-15
[FIX]: Dashboard Tiles Size

### LINKED ISSUE #164: [FIX]: Block Tiles Size
## Summary

Could we make that the tiles grow with the page? That they always have lets say 24px spacing in between. 
You could make the black background only as a color not an image, and use the color "blob" as image and position it relative to the black box then in top-right, bottom, right corner..."

### Suggestions to fix

1. Use `display: grid` instead `display: flex` on container;
   - **Why?** because with `display: grid` you can set the ammount of columns you need (for example: `grid-template-columns: repeat(3, 1fr)` - this will do 3 columns with the same size).
1. Use only the blob as `svg`, and create the card manually.
   - Position manually the blob using `position: absolute`.

## User story it relates to

#118 

### Basic example

image

dsousa12:
Fixes #164 

image

## Proposed Changes

  - Use `display: grid` on container instead `display: flex`
     - **Why?** with `display: grid` you can set the amount of columns you need (`grid-template-columns: repeat(3, 1fr)`)
  - Use _blob_ as a icon, creating the card manually (setting background, etc..)
  
This PR closes #164


nunocaseiro (approved):
LGTM 🚀

PR #211, opened 2022-05-27
[FIX]: Wrong Feedback on first login attempt 

### LINKED ISSUE #209: [FIX]: Wrong Feedback on first login attempt
## Summary

When user try make your first login attempt with wrong data (email or password), appears a "Connection Error" instead of "Credential Error".

## User story it relates to

#14

dsousa12:
Relates to #209

## Proposed Changes

  - Create a `useEffect` to avoid show the default message from enum
     - Basically, when toastr is open, the code sended to get the message from enum is `-1`, returning the default message from that enum.
  
This pull request closes #209

### DIFF
--- a/frontend/components/auth/LoginForm/index.tsx
+++ b/frontend/components/auth/LoginForm/index.tsx
@@ -2,7 +2,7 @@ import { zodResolver } from '@hookform/resolvers/zod';
 import router from 'next/router';
 import { RedirectableProviderType } from 'next-auth/providers';
 import { signIn } from 'next-auth/react';
-import { Dispatch, SetStateAction, useState } from 'react';
+import { Dispatch, SetStateAction, useEffect, useState } from 'react';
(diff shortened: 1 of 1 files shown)


dsousa12:
## **Solution**

The toastr message is an enum and is used a `useState` to "save" the response error and get a message from enum.
The problem is, when show the toastr, this state contain the value `-1`. So the best approach it's create an `useEffect` to show the toastr (only when have a change on this state)


```ts
	useEffect(() => {
		if (loginErrorCode !== -1) {
			setToastState({
				open: true,
				type: ToastStateEnum.ERROR,
				content: getAuthError(loginErrorCode)
			});
		}
		// eslint-disable-next-line react-hooks/exhaustive-deps
	}, [loginErrorCode]);
```

> **Note:** the `eslint-disable-next-line` it's only to garantee that `useEffect` only runs for the `loginErrorCode` state

RubenMCCarreira (approved):

PR #217, opened 2022-05-30
[FEATURE]: Remove order select and add a “sort by votes” button on each column

### LINKED ISSUE #215: [FEATURE]: Remove order select and add a “sort by votes” button on each column
## Summary

On boards page, remove the **Order** button and add a icon to sort by votes each column.

[Figma - Design](https://www.figma.com/file/McDBByDqFzjYuYZ4CwVpax/?node-id=1485%3A138538)

## User story it relates to

#144

dsousa12:
Relates to #215 

image
image

## Proposed Changes

  - Create a popover menu to show the sort by votes options
  - Remove unused filter props
  
This pull request closes #215


dsousa12:
## **Note:**

```ts
	let icon;

	switch (filter) {
		case 'asc':
			icon = 'sort_upvoted';
			break;
		case 'desc':
			icon = 'sort_downvoted';
			break;
		default:
			icon = 'sort';
			break;
	}
```

I use `switch` `case` on the code above, beucase when I try to use a ternary, the `eslint` show an error and the solution is disable `eslint` to that line. But I prefer use another solution, and choose the `switch` `case` alternative.

frontend/pages/boards/[boardId].tsx
```diff
@@ -164,39 +157,26 @@ const Board: React.FC<BoardProps> = ({ boardId, mainBoardId }) => {
...
+						{/*			value: filter, */}
+						{/*			label: `${filter.charAt(0).toUpperCase()}${filter.substring(1)}` */}
+						{/*		}} */}
+						{/*		options={[ */}
+						{/*			{ value: 'order', label: 'Order' }, */}
+						{/*			{ value: 'votes', label: 'Votes' } */}
+						{/*		]} */}
+						{/*		onChange={(option) => setFilter((option as OptionType)?.value)} */}
+						{/*	/> */}
+						{/* </Flex> */}
```
Comment (RubenMCCarreira) [inline frontend/pages/boards/[boardId].tsx:179]: if this code is not used, pls remove it

frontend/types/column.ts
```diff
@@ -27,7 +27,7 @@ export interface ColumnBoardType {
 	maxVotes?: number;
 	countAllCards: number;
 	isSubmited: boolean;
-	filter: string;
+	// filter: string;
```
RubenMCCarreira, line 30: same here

frontend/components/Board/Column/Column.tsx
```diff
@@ -48,28 +24,56 @@ const Column = React.memo<ColumnBoardType>(
 		boardUser,
 		maxVotes,
 		countAllCards,
-		isSubmited,
-		filter
+		isSubmited
 	}) => {
+		const [filter, setFilter] = useState<'asc' | 'desc' | undefined>();
+
+		const filteredCards = () => {
```
tiagomichaelsousa, line 31: we can do this in the FE but it would be better to have a query string filter from the BE 👀

dsousa12: But the filter is for each column, so you can have like sort by votes (asc) and in the Actions column have sort by votes (desc) for example.
But I don't know, let me know what you think and how I can improve it!

RubenMCCarreira (dismissed):

tiagomichaelsousa (approved):

PR #219, opened 2022-05-31
[SUGGESTION]: Use `src/` folder on frontend

### LINKED ISSUE #218: [SUGGESTION]: Use `src/` folder on frontend
## Summary

- Create a `src` folder;
- Set `baseUrl` on `tsconfig.json` to have a cleanner imports;

### Motivation

Avoid have a lot of folders and files on root of frontend and to make the imports cleaner (without a bunch of `../`)

dsousa12:
Relates to #218 

## Proposed Changes

  - Create `src/` folder;
  - Set `baseUrl` as `src/` on `tsconfig` to clean imports;
  - Update import organizer and other configs on `.eslintrc`

### Folder Structure on root  
  ```
  .
├── cypress
├── docker
├── node_modules
├── public
└── src
  ```
  
This pull request closes #218


tiagomichaelsousa:
The idea here is to avoid long path imports? If so, instead of changing the entire folder structure for the project I would suggest creating aliases like `"@hooks": ["src/hooks"]` and them like:

```js
import useUser from '@hooks/useUser';
```

dsousa12:
> The idea here is to avoid long path imports? If so, instead of changing the entire folder structure for the project I would suggest creating aliases like `"@hooks": ["src/hooks"]` and them like:
> 
> ```js
> import useUser from '@hooks/useUser';
> ```

It's an alternative.
I personally prefer using the `src/` folder to avoid a longer root folder, but if you prefer that approach I can close this pull request, and create a new one with aliases

tiagomichaelsousa:
> > The idea here is to avoid long path imports? If so, instead of changing the entire folder structure for the project I would suggest creating aliases like `"@hooks": ["src/hooks"]` and them like:
> > ```js
> > import useUser from '@hooks/useUser';
> > ```
> 
> It's an alternative. I personally prefer using the `src/` folder to avoid a longer root folder, but if you prefer that approach I can close this pull request, and create a new one with aliases

sorry I didn't notice that we don't have an src folder! Im totally up for it as it keeps the root cleaner 🧹
we can still use both approaches and add the aliases but let's get things separated and create another issue/pr with it

tiagomichaelsousa (approved):

PR #259, opened 2022-06-24
[FIX]: Handling the minimum of users error

### LINKED ISSUE #254: [FIX]: Handling the minimum of users error
## Description

When you make the first login/register in the app and try create a new board, you need to have a minimum of 4 users.

image

### Steps to reproduce

1. Register/Login in the app;
2. Click on **Create board** button (on Dashboard or Boards page);

### Expected result

Show the error alert as appears to the team.

[Design](https://www.figma.com/file/McDBByDqFzjYuYZ4CwVpax/UI-v2.0?node-id=1485%3A129850)

- [x] Alert already created 

### Actual result

Appears an error from Next.js

## Idea to solution

Use the `createBoardError` atom to catch the error and set to true, this way, it's appear a alert in the create board page.

> **Note:** alert already created on `[boardId].ts` page, only need to set this state for this particularly error.

### Use atom Value
```ts
```

### Use atom state
```ts
const [haveError, setHaveError] = useRecoilState(createBoardError);
```
> I use this to set the error on `TeamSubTemasConfigurations.tsx` component.

joaosantos99:
Fixes #254

Screenshot 2022-06-24 at 16 49 46

## Proposed Changes

  - Added the error alert when when the team have less than 4 members
  - Stopped the creation of multiple boards ( cause of the issue error ) when the team have less than 4 members

This pull request closes #254

### DIFF
--- a/frontend/src/components/CreateBoard/SubTeamsTab/TeamSubTeamsConfigurations.tsx
+++ b/frontend/src/components/CreateBoard/SubTeamsTab/TeamSubTeamsConfigurations.tsx
@@ -30,6 +30,7 @@ const TeamSubTeamsConfigurations: React.FC = () => {
 	const setBoardData = useSetRecoilState<CreateBoardData>(createBoardDataState);
 	const setToastState = useSetRecoilState(toastState);
 	const [haveError, setHaveError] = useRecoilState(createBoardError);
+	const minMembers = 4;
 
 	/**
 	 * States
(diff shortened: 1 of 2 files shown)


nunocaseiro (approved):

frontend/src/components/CreateBoard/SubTeamsTab/TeamSubTeamsConfigurations.tsx
```diff
@@ -30,6 +30,7 @@ const TeamSubTeamsConfigurations: React.FC = () => {
 	const setBoardData = useSetRecoilState<CreateBoardData>(createBoardDataState);
 	const setToastState = useSetRecoilState(toastState);
 	const [haveError, setHaveError] = useRecoilState(createBoardError);
+	const minMembers = 4;
```
dsousa12, line 33: Not a problem, but as `minMembers` is a constant value maybe declare using uppercase letters (`MIN_MEMBERS`)

frontend/src/hooks/useCreateBoard.tsx
```diff
@@ -15,6 +15,7 @@ const useCreateBoard = (team: Team, stakeHolders: string[]) => {
 	const { board } = createBoardData;
 
 	const minTeams = 2;
+	const minMembers = 4;
```
dsousa12, line 18: Same:
> Not a problem, but as minMembers is a constant value maybe declare using uppercase letters (MIN_MEMBERS)

dsousa12 (changes requested):

frontend/src/components/CreateBoard/SubTeamsTab/TeamSubTeamsConfigurations.tsx
joaosantos99: It makes sense I will proceed to change

dsousa12 (approved):
LGTM 🚀

PR #266, opened 2022-06-27
fix: check user email or unique name in AD

### LINKED ISSUE #265: [BUG]: AD- Use unique email if email is undefined
## Description

If the email prop from auth token is undefined, the unique name should be used to check if the email exists in the ad

nunocaseiro:
Fixes #265 

## Proposed Changes

  - Send the unique name or the email to check if the user exists in ad

This pull request closes #265


dsousa12 (approved):
LGTM 🚀

PR #267, opened 2022-06-27
[FEATURE]: Migrate from Zod to Joi

### LINKED ISSUE #264: [FEATURE]: Migrate from Zod to Joi
## Summary

Use [**Joi**](https://www.npmjs.com/package/joi) as Schema Validation package on frontend

### Motivation

The **backend** already use **Joi**, so with this migration the project will use the same package on backend and frontend

dsousa12:
Relates to #264 

## Proposed Changes

As discussed on Slack, **Backend**  already uses **Joi** as schema validator, so in this pull request I changed all schemas from **Zod** to **Joi** and change on **React Hook Form** the `zodResolver` by `joiResolver`.

  - Rename all schemas files from `.tsx` to `.ts` (don't have any React code inside this files);
  - Install **Joi**;
  - Remove **Zod**;
  - Update all schemas and usage on **React Hook Form**

This pull request closes #264


frontend/src/schema/schemaAddCommentForm.ts
```diff
@@ -0,0 +1,11 @@
+import Joi from 'joi';
+
+const SchemaAddCommentForm = Joi.object({
+	text: Joi.string().min(1).required().messages({
+		'any.required': 'Please write your comment',
+		'string.empty': 'Please write your comment',
+		'string.min': 'Your comment needs at least 1 characters'
```
tiagomichaelsousa, line 7: we don't have i18n configured?

frontend/src/components/CardBoard/CardAvatars.tsx
```diff
@@ -79,24 +71,18 @@ const CardAvatars = React.memo<CardAvatarProps>(
...
-			return col;
-		}, [getRandomColor, stakeholdersColors, stakeholders]);
-
 		const renderAvatar = useCallback(
-			(value: User | string, idx) => {
+			(value: User | string, avatarColor, idx) => {
 				if (typeof value === 'string') {
 					return (
 						<Avatar
 							key={`${value}-${idx}-${Math.random()}`}
```
tiagomichaelsousa, line 79: Out of context of the PR but is the `value` just a string? It's typed as a `User` but then we use the variable as a string, shouldn't it be an object? 

Also, for the avatar shouldn't the `user.id` + `card.id` be enough instead of using idx and `Math.random()`?

frontend/src/components/CardBoard/CardAvatars.tsx
```diff
@@ -108,25 +94,38 @@ const CardAvatars = React.memo<CardAvatarProps>(
 					>
 						<div>
```
tiagomichaelsousa, line 95: Do we need this empty `<div />`?

frontend/src/schema/schemaAddCommentForm.ts
dsousa12: No, but I'm thinking, why not create a common messages.
 What you think?

frontend/src/components/CardBoard/CardAvatars.tsx
dsousa12: The strings used when [appears this error](https://www.figma.com/file/McDBByDqFzjYuYZ4CwVpax/UI-v2.0?node-id=1485%3A129850).
So we need to fake the users.

frontend/src/components/CardBoard/CardAvatars.tsx
dsousa12: No Idea, but I will see if works without this `<div>`.

dsousa12: @tiagomichaelsousa so I try remove the `<div>` and the tooltip not working on `<Avatar>` component without that `<div>`, so this is the reasone why have a empty `<div>`.

frontend/src/hooks/useAvatarColor.tsx
```diff
@@ -0,0 +1,38 @@
...
+			const user = JSON.parse(userStorage);
+			setColor({
+				bg: user.bg,
+				fontColor: user.fontColor
+			});
+		} else {
+			const newColor = isDefaultColor ? getDefaultColor : getRandomColor();
+			localStorage.setItem(localStorageKey, JSON.stringify(newColor));
+			setColor(newColor);
+		}
```
RubenMCCarreira, line 32:
```suggestion
                         return;
		}
		
		const newColor = isDefaultColor ? getDefaultColor : getRandomColor();
		localStorage.setItem(localStorageKey, JSON.stringify(newColor));
		setColor(newColor);
```

dvpfran (approved):

PR #268, opened 2022-06-27
[FIX]: Deploy frontend error

### LINKED ISSUE #254: [FIX]: Handling the minimum of users error

joaosantos99:
Relates to #259 
Fixes #254

## Proposed Changes

  - Add an operation to see if the "team?.users.length" is not undefined before compare with the minimum number of members.

### DIFF
--- a/frontend/src/components/CreateBoard/SubTeamsTab/TeamSubTeamsConfigurations.tsx
+++ b/frontend/src/components/CreateBoard/SubTeamsTab/TeamSubTeamsConfigurations.tsx
@@ -53,7 +53,7 @@ const TeamSubTeamsConfigurations: React.FC = () => {
 	 * If no, redirect to previous router and show a toastr
 	 */
 	useEffect(() => {
-		if ((data && !data[0]) || team?.users.length < MIN_MEMBERS) {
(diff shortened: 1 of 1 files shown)


nunocaseiro (approved):
LGTM 🚀

PR #252, opened 2022-06-24
[FIX]: Docker Configs

dsousa12:
Fixes #251 

## Proposed Changes

  - Change the `.env` variables
  - Update `docker-compose` and `Dockerfile` (backend) file

This pull request closes #251

PR #286, opened 2022-06-29
[FIX]: Pipeline Error

dsousa12:
Fixes #252 

## Proposed Changes

Fix pipeline error after merge #252:
```
./src/components/auth/LoginForm/index.tsx:40:33
Type error: Expected 0 arguments, but got 1.
 38 | const setToastState = useSetRecoilState(toastState);
 39 | const [loginErrorCode, setLoginErrorCode] = useState(-1);
> 40 | const { loginAzure } = useUser(setLoginErrorCode);
    | ^
 41 | const methods = useForm<LoginUser>({
 42 | 		mode:'onBlur',
 43 | 		reValidateMode:'onBlur',
```


dvpfran:
The [SignUpOptionsForm.tsx](https://github.com/xgeekshq/split/blob/main/frontend/src/components/auth/SignUp/SignUpOptionsForm.tsx) file is causing the same error.

PR #308, opened 2022-07-06
[FIX]: Extra votes left on card

### LINKED ISSUE #304: [FIX]: Extra votes left on card
## Summary

If we have a limit of 10 votes and we change to 6, the votes that we have used will remain on the cards, even if it surpasses the new limit. Also it makes the toast appear with negative numbers.

### Basic example

Screenshot 2022-07-04 at 14 54 31

joaosantos99:
Fixes #304

## Proposed Changes

  - The user can't input a number lower the actual maxVotes

This pull request closes #304

### DIFF
--- a/backend/src/modules/boards/dto/board.dto.ts
+++ b/backend/src/modules/boards/dto/board.dto.ts
@@ -7,7 +7,6 @@ import {
   IsOptional,
   IsBoolean,
   IsMongoId,
-  IsNumber,
   Validate,
 } from 'class-validator';
 import { Transform, TransformFnParams, Type } from 'class-transformer';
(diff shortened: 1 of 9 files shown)


dvpfran:
Your solution is working. I'm just thinking if it doesn't make sense to allow decreasing the number of votes while the board still doesn't contain votes.

dsousa12:
> Your solution is working. I'm just thinking if it doesn't make sense to allow decreasing the number of votes while the board still doesn't contain votes. 

Agree!

joaosantos99:
Okk, i will work on that

RubenMCCarreira (approved):

PR #309, opened 2022-07-06
[FEATURE]: hide votes from others on the board

### LINKED ISSUE #236: [FEATURE]: Hide votes from others on the board
### Motivation

Making votes hidden if the Stackholder/SuperAdmin believe it's better for the board not to be biased.

mourabraz:
Relates to #236 

## Proposed Changes

-  When fetch a board check if votes from others users  should be removed from response.

@dsousa12  @CatiaBarroco-xgeeks  @f-morgado 

This pull request closes #236


backend/src/modules/boards/services/get.board.service.ts
```diff
@@ -230,4 +247,47 @@ export default class GetBoardServiceImpl implements GetBoardServiceInterface {
       ],
     });
   }
+
+  private cleanVotes(
+    userId: string,
+    input: LeanDocument<Board & Document<any, any, any> & { _id: any }>,
+  ): LeanDocument<Board & Document<any, any, any> & { _id: any }> {
```
dvpfran, line 254: What do you think change``Board & Document<any, any, any>`` to ``BoardDocument``?
And the ``_id`` we can specify the type to ObjectId: ``{ _id: ObjectId }``

dsousa12: I think if you use only `LeanDocument<Board>` its working, but I don't have sure.

mourabraz: Use `LeanDocument<Board>` is not enough, but I remove all any type inside this function.

dvpfran (approved):

PR #317, opened 2022-07-08
[FIX]: Missing upper case validation on register user

### LINKED ISSUE #311: [FIX]: Missing upper case validation on register user
## Description

We can create a user with all lower cases password, resulting in the login showing the following error:

175A28CD-0247-4A3E-8486-A0545A6D639E

### Steps to reproduce

1. Create a user with all lower cases password.
2. Try to login with the user.

## Expected result

The register of the user should validate the upper case rule.

### Actual result

The register proceed with success resulting on login failure.

joaosantos99:
Fixes #311

## Proposed Changes

  - Changed the regex as mention in the issue

This pull request closes #311


dsousa12 (approved):
LGTM 🚀

PR #339, opened 2022-07-19
[FIX]: Only show the board settings for the user that created the board

### LINKED ISSUE #338: [FIX]: Only show the board settings for the user that created the board
# Description

When a user access the board page, he can change the value of the settings, but when click on save appears an error (because the validation on backend).

So only show the board settings for the user that created the board.

### Steps to reproduce

1. Login into the application
2. Access a board from dashboard or boards page
3. Click on **"Board Settings"** button
4. Change any value ("Hide Votes from others", "Board Name", ...)
5. Click on "Save" button

### Expected result

The user can click on **"Board Settings"** button, if this user create the board. If, the user can't see this button.

### Actual result

Every user can see and click on this button, but when save the **"Board Settings"** appears an error (from Backend validation)

CatiaBarroco-xgeeks:
Fixes #338 

imagem

imagem

## Proposed Changes
  - Only show the button if the user created the board

This pull request closes #338

### DIFF
--- a/frontend/src/pages/boards/[boardId].tsx
+++ b/frontend/src/pages/boards/[boardId].tsx
@@ -71,6 +71,7 @@ const Board: React.FC<BoardProps> = ({ boardId, mainBoardId }) => {
 	});
 	const { data } = fetchBoard;
 	const board = data?.board;
+	const boardOwner = board?.createdBy === userId;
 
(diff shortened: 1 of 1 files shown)


dsousa12 (approved):
LGTM 🚀

PR #360, opened 2022-07-22
[BUG]: Team configuration is reseted when clicking on configurations and coming back to it

### LINKED ISSUE #350: [BUG]: Team configuration is reset when clicking configurations and coming back to it
## Description

When creating a split board, clicking on **Configurations** and then going back to **Team/-Sub-teams configurations** resets the teams

Screenshot 2022-07-22 at 13 58 03

### Steps to reproduce

* Setup teams on the **Team/-Sub-teams configurations**
* Click on **Configurations**
* Click back on **Team/-Sub-teams configurations**

### Expected result

Team configuration should remain the same

### Actual result

The configuration is lost

joaosantos99:
Fixes #350

## Proposed Changes

  - Created a state for identifying when the sub-boards don't need to be reset.

This pull request closes #350

### DIFF
--- a/frontend/src/components/CreateBoard/SubTeamsTab/MainBoardCard.tsx
+++ b/frontend/src/components/CreateBoard/SubTeamsTab/MainBoardCard.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect } from 'react';
-import { SetterOrUpdater, useRecoilState, useRecoilValue } from 'recoil';
+import { SetterOrUpdater, useRecoilValue } from 'recoil';
 
 import { styled } from 'styles/stitches/stitches.config';
 
@@ -18,7 +18,6 @@ import { Team } from 'types/team/team';
 import { BoardUserRoles } from 'utils/enums/board.user.roles';
 import { TeamUserRoles } from '../../../utils/enums/team.user.roles';
 import SubCardBoard from './SubCardBoard';
-import { newBoardState } from 'store/createBoard/atoms/create-board.atom';
 
 const MainContainer = styled(Flex, Box, {
 	backgroundColor: 'white',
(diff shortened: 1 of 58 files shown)


frontend/src/components/CreateBoard/SubTeamsTab/MainBoardCard.tsx
```diff
@@ -57,6 +58,7 @@ const MainBoardCard = React.memo(({ team }: MainBoardCardInterface) => {
 	 * Recoil Atoms
 	 */
 	const haveError = useRecoilValue(createBoardError);
+	const [newBoard, setNewBoard] = useRecoilState(newBoardState);
```
dsousa12, line 61: Why you need a new **Recoil** state to storage a boolean?

dsousa12 (changes requested):

frontend/src/components/CreateBoard/SubTeamsTab/MainBoardCard.tsx
joaosantos99: Because when the user changes from the teams configuration to the board configuration and switch back again to the teams configuration, it re-renders the component and in the useEffect resets any change on the teams. So i added a booleon to know when to reset the teams cahnges.

dsousa12: Ok, but you can't avoid the re-renders without another state?

dsousa12: If not, maybe create a normal `useState` and send by props the `set` function.
Because you don't use this recoil state in other places I think.

But wait for other reviews.

joaosantos99: I initially i used `useState`, but later i needed a way to reset the state when someone clicked on the "Add new board" button so i switched to the atom.
Do you have any suggestion to solve that problem?

dsousa12 (dismissed):
LGTM 🚀

dsousa12 (approved):

PR #374, opened 2022-07-26
[FEATURE]: Change responsible on board settings

### LINKED ISSUE #369: [FEATURE]: Change responsible on board settings
## Summary

list all members in the Settings board and be able to change the responsible

### Motivation

have the possibility to change a person from responsible to member and vice versa.
In this way, when there is someone who cannot be responsible, the situation can be resolved by naming another person as responsible.

CatiaBarroco-xgeeks:
Relates to #369 

imagem

imagem

## Proposed Changes

  - added the possibility to change the responsibility of the board after it has been created
  - CSS changed according to final version

@dsousa12 @f-morgado 

This pull request closes #369


frontend/src/components/Board/Settings/index.tsx
```diff
@@ -317,17 +372,114 @@ const BoardSettings = ({ isOpen, setIsOpen, socketId }: Props) => {
...
+													</Text>
+
+													<Avatar
+														css={{ position: 'relative' }}
+														size={32}
+														colors={{
+															bg: highlight2Colors.highlight2Lighter,
+															fontColor:
+																highlight2Colors.highlight2Dark
+														}}
```
dsousa12, line 467: You try use without import the color and use it like stitches way?

Something like:
```suggestion
														colors={{
															bg: '$highlight2Lighter',
															fontColor: '$highlight2Dark'
'														}}
```

frontend/src/components/Board/Settings/index.tsx
```diff
@@ -4,17 +4,24 @@ import { useRecoilState, useRecoilValue } from 'recoil';
 import { joiResolver } from '@hookform/resolvers/joi';
 import { Accordion } from '@radix-ui/react-accordion';
 import { Dialog, DialogClose, DialogTrigger } from '@radix-ui/react-dialog';
+import { deepClone } from 'fast-json-patch';
+
+import { highlight2Colors } from 'styles/stitches/partials/colors/highlight2.colors';
```
dsousa12, line 9: If the comment I leave bellow works, please remove this import.

frontend/src/components/Board/Settings/index.tsx
dsousa12: Already tested it and works without import `highlight2Colors`.

dsousa12 (changes requested):

backend/src/modules/boards/services/update.board.service.ts
```diff
@@ -64,6 +74,18 @@ export default class UpdateBoardServiceImpl implements UpdateBoardService {
...
+	/**
+	 * Method to get current responsible to a specific board
+	 * @param boardId String
+	 * @return Board User
+	 * @private
+	 */
+	private async getBoardResponsible(boardId: string): Promise<BoardUser | undefined> {
+		const users = await this.boardUserModel.find({ board: boardId }).exec();
+
+		return users.find((user) => user.role === BoardRoles.RESPONSIBLE);
```
geomarb, line 86: There is a way to get the result in the first `find` without the need of the second find, you can find an example [here](https://www.mongodb.com/docs/manual/reference/method/db.collection.find/ ).
PS: This is a small and insignificant issue that can be disregarded, but could be done more for learning purposes.

dsousa12: Thanks @geomarb, I don't remember that.
Already done, thanks again 😉

backend/src/modules/boards/services/update.board.service.ts
```diff
@@ -89,6 +117,36 @@ export default class UpdateBoardServiceImpl implements UpdateBoardService {
 		const isOwner = String(userId) === String(createdBy);
 
 		if (isAdminOrStakeholder || isOwner || (isSubBoard && !!subBoardResponsible)) {
+			if (boardData.users && isSubBoard) {
```
geomarb, line 120: It would be nice to have some comments in this part explaining what it does.

dsousa12: Agree, but basically to you understand:
1. First validate if the user have permissions to edit this specific board
2. If have permissions, validate if have an array of users (only have when change responsible) and is a sub-board.
    - **Note:** only on sub-boards have the **responsible** role, so needs to check if is a sub-board.

frontend/src/components/Board/Settings/index.tsx
```diff
@@ -154,27 +214,56 @@ const BoardSettings = ({ isOpen, setIsOpen, socketId }: Props) => {
 	};
 
 	const updateBoard = (title: string, maxVotes: string | undefined) => {
-		mutate(
-			{
+		console.log('aqi', updateBoardData);
+		mutate({
```
geomarb, line 218: Please, remove this console.log.

geomarb (changes requested):
There are some small issues to be reviewed such as a the rule to avoid changing maxVotes and the getResponsible function.

backend/src/modules/boards/services/update.board.service.ts
```diff
@@ -48,20 +59,48 @@ export default class UpdateBoardServiceImpl implements UpdateBoardService {
 	 * If not, return a null value
 	 *
 	 * @param userId Current User Logged
+	 * @param boardId Board Id
 	 * @returns Board User
 	 */
-	private async getResponsible(userId: string): Promise<LeanDocument<BoardUserDocument> | null> {
-		const user = await this.boardUserModel.findOne({ user: userId }).lean().exec();
+	private async checkIfUserIsResponsible(userId: string, boardId: string): Promise<boolean> {
```
geomarb, line 65: It is ok like it is. A shorter version could be `isUserResponsiple(...)`.

backend/src/modules/boards/services/update.board.service.ts
```diff
@@ -83,23 +122,77 @@ export default class UpdateBoardServiceImpl implements UpdateBoardService {
...
+				const currentResponsibleId = await this.getBoardResponsibleId(boardId);
+				const newResponsibleId = (
+					boardData.users.find((user) => user.role === BoardRoles.RESPONSIBLE)
+						?.user as unknown as LeanDocument<BoardUserDocument>
+				)._id;
+
+				boardData.users
+					.filter((boardUser) =>
+						[
+							replaceAll(String(currentResponsibleId), ['"', '(', ')', 'ObjectId'], ''),
```
geomarb, line 147: It seems we can get the currentResponsibleId without the help of replaceAll (which could be dropped if it is not used somewhere else).

backend/src/libs/utils/replace-all.ts
```diff
@@ -0,0 +1,11 @@
+const replaceAll = (value: string, find: string[], replace: string): string => {
```
geomarb, line 1: If change the way to get the userId directly instead of an object, we could remove this function.

geomarb (changes requested):
LGTM, but it has some small adjusts that could be done, I put a comment the in the lines where I believe could be different. 
We need check with the designer about the way we change the stakeholder as a user, since we need click three times and the lottery button doesn't looks like a button what is not so intuitive.
Talking about the design/user experience it would be better if we could keep Save button deactivated until we have any changes.

backend/src/modules/boards/services/update.board.service.ts
dsousa12: I create this function because the other two IDs return only the id, but the `currentResponsibleId` returns with `ObjectId('user-id')`.
So to prevent any error, I made this function to replace the `ObjectId`, `()` and `"`

geomarb: Got it now thanks! As it has a very specific purpose it would be better to rename `replaceAll()` to something like 
`getIdFromObjectId(objectId: ObjectId) {...}` and inside of this function do all the job needed as converting ObjectId to string and the array need to clean data inside the function.

backend/src/libs/utils/replace-all.ts
geomarb: I am setting this request as resolved as it works, in this way we don't need to hold this PR. This change is optional.

backend/src/modules/boards/services/update.board.service.ts
geomarb: I set it as resolved as this change is optional.

dsousa12:
## Note:

To avoid errors related with event listener created to use when user press key, close the dialog and submit the form, we need create a `span`. around the dialog.
This `span`. contains a `ref={}` and the event listener is used only on the content of this `span`.

dsousa12 (approved):

PR #336, opened 2022-07-18
[FEATURE]: Create slack channels

mourabraz:
Relates to #153 

## Proposed Changes

  - Add new feature to create channels and invite users based on a board

This pull request closes #153

PR #375, opened 2022-07-26
[FEATURE]: Call communication execute with queue

mourabraz:
Relates to #336 

## Proposed Changes

  - add queue module with bull and Redis
  - change execute method in communication service to call the producer to slack queue
  - add unit tests for producer and consumer


backend/src/modules/communication/producers/slack-communication.producer.service.ts
```diff
@@ -0,0 +1,44 @@
...
+	constructor(
+		@InjectQueue(CommunicationProducerService.QUEUE_NAME)
+		private readonly queue: Queue
+	) {
+		// https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#events
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		this.queue.on('completed', (job: Job<JobType>, data: TeamDto[]) => {
+			this.logger.verbose(`Completed Job id: "${job.id}"`);
+			// TODO
+			// console.log(data);
```
tiagomichaelsousa, line 23: this can be removed i guess

backend/src/modules/queue/queue.module.ts
```diff
@@ -0,0 +1,32 @@
...
+				redis: {
+					username: configService.get('redis.user'),
+					password: configService.get('redis.password'),
+					host: configService.get('redis.host'),
+					port: configService.get('redis.port'),
+					tls: {
+						host: configService.get('redis.host')
+					}
+				},
+				name: CommunicationProducerService.QUEUE_NAME,
```
tiagomichaelsousa, line 23: I don't think that the bull/queue module should be the one to register the queues.  Instead, the modules should import the bull/queue config and register themselves.
For example:

```js
// backend/src/modules/queue/queue.module.ts
@Module({
  imports: [
    BullModule.forRootAsync({
      inject: [ConfigService],
      useFactory: (config: ConfigService) => ({
        redis: {
          username: configService.get('redis.user'),
          password: configService.get('redis.password'),
          host: configService.get('redis.host'),
          port: configService.get('redis.port'),
          tls: {
            host: configService.get('redis.host')
          }
        },
      }),
    }),
  ],
})
```
And then, for the module that will use the queue we can do something like this:
```js
// communication.module.ts
@Module({
  imports: [BoardsModule, QueueModule],
  providers: [
    SlackExecuteCommunicationService,
    CommunicationGateAdapter,
    ChatHandler,
    ConversationsHandler,
    UsersHandler,
    ExecuteCommunication
  ],
  exports: [SlackExecuteCommunicationService],
  imports: [
    BullModule.registerQueueAsync({
      name: 'slack-queue', // this can be a constant in the CommunicationProducerService.QUEUE_NAME
    }),
  ],
})
```

With that, every module that will require/need queues can import the bull configuration and register themselves async in the queue system, otherwise, your queue configuration can get pretty huge depending on how many modules you have that will need a queue system 🙂

Also, o do think that `Communication` is a very broad naming for the slack integration, could be more specific to the use case and call it `SlackModule`, `SlackProducer` and `SlackService` but no need to change it for this PR as it changes a lot the scope of the task

backend/src/modules/communication/producers/slack-communication.producer.service.ts
```diff
@@ -0,0 +1,44 @@
...
+			job.remove();
+		});
+		this.queue.on('error', (error: Error) => {
+			this.logger.error(`Error on queue: "${JSON.stringify(error)}"`);
+		});
+		this.queue.on('failed', (job: Job<JobType>, error) => {
+			this.logger.error(`Job id: "${job.id}" fails with error: "${JSON.stringify(error)}"`);
+		});
```
tiagomichaelsousa, line 31: nice job, very important to keep track of what happened in the logs!

backend/src/modules/communication/consumers/slack-consumer.processor.ts
```diff
@@ -0,0 +1,36 @@
...
+import { UsersSlackHandler } from 'modules/communication/handlers/users-slack.handler';
+
+const logger = new Logger('CommunicationConsumerProcessor');
+
+// If any error occurs the error will propagate and will be captured on queue's events
+// it is not necessary use the "cb" to deal with the Errors
+export default async function consumer(job: Job<JobType>, cb: DoneCallback) {
+	const data = job.data;
+	const board = data.board;
+	const config: ConfigurationType = data.config;
```
tiagomichaelsousa, line 18:
```js
const { data, board, config } = data;
```

tiagomichaelsousa (commented):
Awesome work, left a couple of comments

backend/src/modules/communication/producers/slack-communication.producer.service.ts
mourabraz: removed.

backend/src/modules/queue/queue.module.ts
mourabraz: moved queue registration to communication module.
But as you mention I did not change the names for now.

backend/src/modules/communication/consumers/slack-consumer.processor.ts
mourabraz: changed to `const { board, config } = job.data;`

backend/src/modules/queue/queue.module.ts
mourabraz: Taking this change in consideration I moved the QueueModule to `app.module.ts`. For now it is being added side by side with `CommunicationModule` because this is the only one using it, and with QueueModule some environment variables for Redis will be mandatory.  I want to avoid this if slack is not enable.

f-morgado (approved):
LGTM 🚀

PR #385, opened 2022-07-29
[FEATURE]: Hide id from users when they vote on the cards

### LINKED ISSUE #378: [FEATURE]: Hide id from users when they vote on the cards
## Summary

Users' votes should not show the id of whoever made the vote

### Motivation

Making the votes anonymous and not traceable

f-morgado:
Relates to #378 
## Proposed Changes

  - Users' id is hidden when he votes so that other can't see who placed votes on the card (beside themselves)

@CatiaBarroco-xgeeks @dsousa12 

This pull request closes #378

### DIFF
--- a/backend/src/modules/boards/services/get.board.service.ts
+++ b/backend/src/modules/boards/services/get.board.service.ts
@@ -257,10 +257,18 @@ export default class GetBoardServiceImpl implements GetBoardServiceInterface {
 		userId: string,
 		hideCards: boolean,
 		hideVotes: boolean
+		// votesId: string
 	): LeanDocument<CardDocument | CardItemDocument> {
 		let { text, comments, votes, createdBy } = input;
 		const { anonymous } = input;
 		const createdByAsUserDocument = createdBy as UserDocument;
+
+		// if (String(createdByAsUserDocument._id) !== String(userId))
+		votes.forEach((vote) => {
+			console.log('voteId to replace', String(vote));
+			vote = hideText(String(vote));
+			console.log('voteId replaced', String(vote));
+		});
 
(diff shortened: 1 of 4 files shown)


backend/src/modules/boards/services/get.board.service.ts
```diff
@@ -276,6 +278,8 @@ export default class GetBoardServiceImpl implements GetBoardServiceInterface {
 			votes = this.replaceVotes(input, userId);
 		}
 
+		votes = this.replaceAllVotes(input, userId);
```
mourabraz, line 281: There are no differences between `replaceVotes` and `replaceAllVotes`. 
So `votes` is changing with or without `hideVotes`

backend/src/modules/boards/services/get.board.service.ts
```diff
@@ -285,6 +289,11 @@ export default class GetBoardServiceImpl implements GetBoardServiceInterface {
 		};
 	}
 
+	//-----------------------------
+	private replaceAllVotes(input: LeanDocument<CardDocument | CardItemDocument>, userId: string) {
```
mourabraz, line 293: The name could be something related to `filter` since we are filtering and not replacing.

backend/src/modules/boards/services/get.board.service.ts
f-morgado: True, removed it!

backend/src/modules/boards/services/get.board.service.ts
f-morgado: Changed it to filterVotes ;)

PR #401, opened 2022-08-10
[FEATURE]: Unit tests

### LINKED ISSUE #400: [FEATURE]: Unit tests
## Summary

Adding unit tests to the frontend components.

### Motivation

Mitigate errors before pushing to the main.

joaosantos99:
Relates to #400 

## Proposed Changes

  - Adding unit test to the frontend components

This pull request closes #400

### DIFF
--- a/frontend/src/components/Board/Card/CardBoard.test.tsx
+++ b/frontend/src/components/Board/Card/CardBoard.test.tsx
@@ -4,7 +4,9 @@ import { useRouter } from 'next/router';
 import { DragDropContext, Droppable } from '@react-forked/dnd';
 import { render } from '@testing-library/react';
 
+import { BoardUser } from 'types/board/board.user';
 import CardType from 'types/card/card';
+import { BoardUserRoles } from 'utils/enums/board.user.roles';
 import { CardsContainer } from '../Column/styles';
 import CardBoard from './CardBoard';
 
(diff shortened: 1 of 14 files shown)


frontend/src/components/Board/Card/CardBoard.test.tsx
```diff
@@ -0,0 +1,279 @@
...
+const mockSession = {
+	expires: 10,
+	user: {
+		id: '62f104ffc81c816dd0cf7765',
+		email: 'email@test.com'
+	},
+	isSAdmin: false,
+	joinedAt: '',
+	session: ''
+};
```
dsousa12, line 53: ## **Sugggestion:**
I think it's a better approach if we create a `__fixtures__` folder to store the mock info and reuse into multiple tests

frontend/src/components/Board/Card/CardBoard.test.tsx
```diff
@@ -0,0 +1,279 @@
...
+			_id: '62f104ffc81c816dd0cf7765',
+			firstName: 'FirstName',
+			lastName: 'LastName',
+			email: 'email@test.com',
+			isSAdmin: false,
+			joinedAt: ''
+		},
+		anonymous: true,
+		_id: '62f221dfc81c816dd0cf7825'
+	};
```
dsousa12, line 183: Same here, you can create a `__fixtures__/cards.ts` and store this info, and that way you can use on next test

frontend/src/components/Board/Card/CardBoard.test.tsx
mourabraz: I agree with you when we have factories to create mocked data. But when I use mocks like this I prefer to keep it close. However, as soon as we start to use this mocked data in other tests I prefer to use a factory and mock the data dynamically. (using faker.js for example)

PR #433, opened 2022-09-07
[FEATURE]: Votes request with timer

### LINKED ISSUE #368: [BUG]: vote counter goes negative if users surpasses the limit of votes to use
## Summary

Vote counter limited to 0 and not going negative

### Motivation

Avoiding erros and giving the wrong info about remaining votes with negative numbers.

### LINKED ISSUE #386: [REFACTOR]: Votes Structure
## Summary

Currently each vote represents an entry on array containing the **User ID**, for example:

```
votes: ['user 1', 'user 2', 'user 1']
```

### Problems?
- If an user leave 20 votes in one card, the same ID appears 20 times on that array
- Another problem is you can't use a debouncer (for example) to avoid a lot of requests if the user spam the votes button. **Why?** because the request to add or remove a vote only add a new entry.

### Improvements

If we change the votes structure for something similar to this":
```ts
votes: {
	userId: string,
	votes: number
}[]
```

- We can use a debouncer to avoid the multiple requests if the user spams the vote button, changing the request to send the number of votes to add instead only incrementing the value (for example);
- No need to reference the **User ID** with **User** schema (my opinion);
- The array of votes is cleaner, because the **User ID** only appears one time.

### Motivation

Fix the problem when user spam the add/remove votes button (a lot of requests) and create a more readable structure.

### Tasks

- [ ] Change schemas
	- `card.item.schema.ts`
	- `card.schema.ts`
- [ ] Create new schemas for votes
- [ ] Create new DTOs for votes
- [ ] Change controllers and services
	- `votes.controller.ts`
	- `create.vote.service.ts`
	- `delete.vote.service.ts`
- [ ] Change the logic on frontend
- [ ] **OPTIONAL:** user debouncer and instead increment votes, receive a number of votes on `body`

### LINKED ISSUE #413: [BUG] Upvoting a merged card crashes the website
**Description**
As a user, when I upvote a card that was merged, the website crashes

**Current behavior**
Website crashes upon clicking on the thumbs up button on a merged card:

https://user-images.githubusercontent.com/87712174/185373828-bbeb959c-1602-4450-8882-d4fdced5b46e.mp4

### LINKED ISSUE #415: [BUG]: User votes surpassing the board limits's vote
## Summary

user votes are not respecting the limit of the board and crashing the app

### Motivation

Why are we doing this? What use cases does it support? What is the expected outcome?

Avoiding the app to crash after 1 or more votes surpass the limit

### LINKED ISSUE #427: [FEATURE]: Votes request with timer
## Summary

Users can spam votes and make the requests bug the page due to overlap

### Motivation

Not overlapping several vote requests, instead wait for a certain time before sending a new request

CatiaBarroco-xgeeks:
Fixes #427 
Fixes #415 
Fixes #413 
Fixes #386 
Fixes #368 

imagem

## Proposed Changes
  - add timer on votes request
  - resolve error on surpassing votes
  - merge/unmerge vote fix

@miguel-felix1 @f-morgado @joaosantos99 

This pull request closes #427
This pull request closes #415 
This pull request closes #413 
This pull request closes #386 
This pull request closes #368


backend/src/modules/votes/controller/votes.controller.ts
```diff
@@ -217,4 +219,63 @@ export default class VotesController {
...
+					boardId,
+					cardId,
+					request.user._id,
+					itemId
+				);
+			} else {
+				// eslint-disable-next-line no-await-in-loop
+				board = await this.createVoteApp.addVoteToCard(boardId, cardId, request.user._id, itemId);
+			}
+		}
```
nunocaseiro, line 247: I started by doing this but I think this should be avoided. We need check how many votes to add or remove and do it at once without a calling the `deleteVoteFromCard` or `addVoteToCard` multiple times. Both methods should be modified to receive a count prop. Please check the `fix/vote-limit-frontend` branch I didn't push the last changes

comoser (approved):
go ahead 👍

backend/src/modules/votes/controller/votes.controller.ts
comoser: that's true, this should've been made differently, but for now to close the MVP we can keep the logic working. We need to do many deep refactors to the app. But first, the MVP needs delivery

PR #505, opened 2022-10-18
[FIX]: fix card team layout

### LINKED ISSUE #501: [FIX]: Fix Team card layout
## Summary

Fix the layout of the team card when the user isn't an admin or super admin

### Motivation

The card layout is correct for when the user is an admin but when its a member, the center part goes to the right of the card instead of staying fix in the center

CatiaAntunes96:
Relates to #501 

## Proposed Changes

  - Fix the layout of the team card when the user isn't an admin or super admin

This pull request closes #501


joaosantos99 (approved):
LGTM 🚀

PR #529, opened 2022-10-28
Feat/change searchinput

### LINKED ISSUE #527: [FEATURE]: Change SearchInput
## Summary

Change search input from create new team, from the current input to an input that allows onChange method.
This will allow to filter the members at the same time that the user is typing. 

### Motivation

The input that is being used right now only works with forms and that component needs an input that accepts an onChange event to filter the users list.

CatiaAntunes96:
Relates to #527

## Proposed Changes

  - Search input was changed to an input that accepts an onChange event

This pull request closes #527

### DIFF
--- a/frontend/src/components/Teams/CreateTeam/ListMembers/index.tsx
+++ b/frontend/src/components/Teams/CreateTeam/ListMembers/index.tsx
@@ -1,11 +1,8 @@
-import { Dispatch, SetStateAction, useEffect, useRef } from 'react';
-import { FormProvider, useForm, useWatch } from 'react-hook-form';
-import { joiResolver } from '@hookform/resolvers/joi';
+import React, { Dispatch, SetStateAction, useEffect, useRef, useState } from 'react';
 import { Dialog, DialogClose, DialogTrigger } from '@radix-ui/react-dialog';
 
 import Icon from 'components/icons/Icon';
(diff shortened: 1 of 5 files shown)


frontend/src/components/Teams/CreateTeam/ListMembers/index.tsx
```diff
@@ -99,40 +84,34 @@ const ListMembers = ({ isOpen, setIsOpen }: Props) => {
...
+					<Flex css={{ padding: '$24 $32 $40' }} direction="column" gap={16}>
+						<InputSearch
+							currentValue={searchMember}
+							icon="search"
+							iconPosition="left"
+							id="search"
+							placeholder="Search member"
+							handleChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+								setSearchMember(event.target.value)
+							}
```
nunocaseiro, line 97: I think it is a best practice to use a function reference instead of declaring the function inside the `handle change`
```suggestion
							handleChange={handleSearchChange}
```

```
const handleSearchChange = (event: React.ChangeEvent<HTMLInputElement>) => {
   setSearchMember(event.target.value)
}
```

nunocaseiro (changes requested):

nunocaseiro (approved):
LGTM 🚀

PR #540, opened 2022-11-02
[FEATURE]: list of selected users on recoil

### LINKED ISSUE #530: [FEATURE]: list of selected users on recoil
## Summary

The users that are being selected to be part of a team should be kept on recoil. This will allow access to the list, when the team will be created

### Motivation

This will avoid the use of props between the components and the lifting up the state to the component were the information will be submitted. This way, the components that need access to the members that are being added just need to use recoil.

CatiaAntunes96:
Relates to #530 

## Proposed Changes

  - Add checked users to recoil on Add button
 

This pull request closes #530


nunocaseiro (approved):
LGTM 🚀

PR #542, opened 2022-11-02
[Refactor]: slack schedules and slack option

### LINKED ISSUE #519: [BUG]: Board created is not persisting value of "slackEnable" option
## Description

When a user creates a new board the option for slack is not being persisted. We need this value for the recurrent board.

### Expected result

The persisted data for the new board should have the slack option value ("slackEnable" property)

### Actual result

When the cron job recreates the recurrent board it hasn't the information if it should dispatch the current board to Slack communication service.

### Related to 
#157 
#50

nunocaseiro:
Fixes #519

## Proposed Changes

  - Saves the slack option in the scheduling entity
  - Add a cron job when the user creates a new split retro. The job triggers the creation of a new split board based on the previous board settings
  - Fix mongoose package vulnerability 

This pull request closes #


mourabraz (approved):
🚀  ACK

PR #543, opened 2022-11-02
[FEATURE]: endoint to get the teams of a user

### LINKED ISSUE #541: [FEATURE]: getTeamsOfUser endoint
## Summary

Change endpoint getAllTeams to getTeamsOfUser.

### Motivation

The user should only view his teams, not all of them.

CatiaAntunes96:
Relates to #541 

## Proposed Changes

  - On the page "Teams", the endpoint that now is being used is `getTeamsOfUser()` instead `getAllTeams()` 
  - Add endpoint to teams controller
 

This pull request closes #541

### DIFF
--- a/backend/src/modules/teams/controller/team.controller.ts
+++ b/backend/src/modules/teams/controller/team.controller.ts
@@ -36,6 +36,7 @@ import { BadRequestResponse } from 'libs/swagger/errors/bad-request.swagger';
 import { InternalServerErrorResponse } from 'libs/swagger/errors/internal-server-error.swagger';
 import { UnauthorizedResponse } from 'libs/swagger/errors/unauthorized.swagger';
 
+import { UserParams } from '../../../libs/dto/param/user.params';
 import { CreateTeamDto } from '../dto/crate-team.dto';
 import TeamDto from '../dto/team.dto';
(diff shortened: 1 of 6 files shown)


backend/src/modules/teams/controller/team.controller.ts
```diff
@@ -120,6 +121,25 @@ export default class TeamsController {
...
+	@ApiBadRequestResponse({
+		description: 'Bad Request',
+		type: BadRequestResponse
+	})
+	@ApiInternalServerErrorResponse({
+		description: 'Internal Server Error',
+		type: InternalServerErrorResponse
+	})
+	@Get(':userId')
+	getTeamsOfUser(@Param() { userId }: UserParams) {
```
nunocaseiro, line 139: You don't need to send the `userId` as it can be extracted with the bearer token. We are already doing this in other methods e.g. `getAllBoards`.
The endpoint can be modified to something like `@Get('me')` or `@Get('user')`.
Changing this you can omit the frontend logic to retrieve and send the user id

nunocaseiro (changes requested):

nunocaseiro (approved):
LGTM 🚀 great job

PR #547, opened 2022-11-03
[FEATURE]: render members card

### LINKED ISSUE #544: [FEATURE]: render members card
## Summary

After selecting the members from the checklist, this list should render as cards.

### Motivation
Shown the cards with the different members that will be part of the team.

CatiaAntunes96:
Relates to #544 

image

## Proposed Changes

  - Instead of just keep a list of objects with userId and role, now the object is keeping the whole user and his role (they are needed in the card member)
  -The members list is initialise with the user that is creating the team with the role as 'Admin'. This is being done at the same time that the prop isChecked is being added to each user.  
  -When the members are added, the cards with each user are rendered.

This pull request closes #544


nunocaseiro (approved):
LGTM 🚀

PR #560, opened 2022-11-07
[FEATURE]: show boards members 

### LINKED ISSUE #555: [FEATURE]: show boards members
## Summary

When the user is in his sub-board/main board, when he clicks in the icons with the members, a popover should appear with all the users of that sub-board/board.

### Motivation

With this feature the user can see everyone that belongs to his sub-board/main board.

CatiaAntunes96:
Relates to #555 

## Proposed Changes

  - If the user is in a board or sub-board page, he can click on the avatars icon to see all the users in that board. If he clicks again on an avatar, the list goes back to the default (only 3 avatars)

This pull request closes #555

### DIFF
--- a/frontend/src/components/CardBoard/CardAvatars.tsx
+++ b/frontend/src/components/CardBoard/CardAvatars.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 
 import Avatar from 'components/Primitives/Avatar';
 import Flex from 'components/Primitives/Flex';
@@ -6,6 +6,7 @@ import Tooltip from 'components/Primitives/Tooltip';
 import { User } from 'types/user/user';
 import { BoardUserRoles } from 'utils/enums/board.user.roles';
 import { TeamUserRoles } from 'utils/enums/team.user.roles';
+import { IconButton } from './styles';
 
 type ListUsersType = {
 	user: User | string;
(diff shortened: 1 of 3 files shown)


frontend/src/components/CardBoard/CardAvatars.tsx
```diff
@@ -21,10 +22,26 @@ type CardAvatarProps = {
...
+		haveError,
+		responsible,
+		myBoards,
+		isBoardsPage
+	}) => {
+		const [viewAllUsers, setViewAllUsers] = useState(false);
+
+		const handleViewAllUsers = useCallback(() => {
+			setViewAllUsers(!viewAllUsers);
+		}, [viewAllUsers]);
```
nunocaseiro, line 43:
```suggestion
		const handleViewAllUsers = () => {
			setViewAllUsers(!viewAllUsers);
		};
```

Do you really need this useCallback?

frontend/src/components/CardBoard/CardAvatars.tsx
```diff
@@ -105,13 +149,22 @@ const CardAvatars = React.memo<CardAvatarProps>(
...
 				);
 			},
-			[userId]
+			[handleViewAllUsers, userId, isBoardsPage]
 		);
 
+		const numberOfAvatars = useMemo(() => {
+			if (!myBoards) {
+				if (viewAllUsers && isBoardsPage) return data.length;
+				return 3;
```
nunocaseiro, line 162:
```suggestion
				return viewAllUsers && isBoardPage ? data.length : 3
```

nunocaseiro (changes requested):

frontend/src/components/CardBoard/CardAvatars.tsx
CatiaAntunes96: In the start I had as you suggested but eslint advised me to useCallback because it makes the dependencies of renderAvatar change on every render. But If you think that there is no problem with using your approach I can put an 	`// eslint-disable-next-line react-hooks/exhaustive-deps ` before the function and remove the useCallback

nunocaseiro: No problem, we can continue with your approach

nunocaseiro (approved):

PR #565, opened 2022-11-08
[Fix]: unmerge cards and remaining votes

### LINKED ISSUE #552: [BUG] Stakeholder should not be included in sub-boards
## Description
As a Stakeholder I should have access to the main and sub-boards

### Current behavior
Stakeholders are being included in the sub-board group creation just like any normal user

Image

nunocaseiro:
Fixes #552

## Proposed Changes

  - Fix unmerge optimistic update
  - Fix the cancel button, should return to the previous page - not only to boards list
  - Add remaining votes toast info when the board has a vote limit

This pull request closes #552

### DIFF
--- a/frontend/src/api/boardService.tsx
+++ b/frontend/src/api/boardService.tsx
@@ -197,6 +197,8 @@ export const handleVotes = (voteDto: {
 	isCardGroup: boolean;
 
 	count: number;
+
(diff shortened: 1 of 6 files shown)


CatiaAntunes96 (approved):
LGTM 🚀

PR #629, opened 2022-11-24
chore: change config files and restore original path files

nunocaseiro:
## Proposed Changes

  - Change eslint, jest and ts configurations
  - Restore the path of the files to include `src` to solve some issues
  - Apply some configs that are used in another project


rpvsilva (approved):
LGTM 🚀

PR #633, opened 2022-11-25
feat: add delete team endpoint

### LINKED ISSUE #495: [FEATURE]: Add Delete Team Endpoint
## Summary

Add delete team endpoint on the backend and then add to he frontend

### Motivation

It will allow the super Admin and the team admin to delete the team

patricia-mdias:
Relates to #495 

## Proposed Changes

  - add new endpoint to delete teams

@CatiaAntunes96 

This pull request closes #495


backend/src/modules/teams/services/delete.team.service.ts
```diff
@@ -0,0 +1,41 @@
...
+			{
+				_id: teamId
+			},
+			{ session: teamSession }
+		);
+
+		if (!result) throw new NotFoundException(DELETE_FAILED);
+
+		return { _id: result._id };
+	}
```
nunocaseiro, line 40: Good job here with the transactions. I think you followed the delete board method however, as mongo doesn't have the cascade strategy you should also implement it by deleting manually the teamUsers using the `deleteMany` like we did to the `deleteBoardUsers`. You can check the `delete.board.service` file and do something similar for the teams

nunocaseiro (changes requested):

backend/src/modules/teams/controller/team.controller.ts
```diff
@@ -220,4 +224,34 @@ export default class TeamsController {
...
+	})
+	@TeamUser(TeamRoles.ADMIN)
+	@UseGuards(TeamUserGuard)
+	@Delete(':teamId')
+	async deleteTeam(@Param() { teamId }: TeamParams) {
+		const result = await this.deleteTeamApp.delete(teamId);
+
+		if (!result) throw new BadRequestException(DELETE_FAILED);
+
+		return result;
```
rpvsilva, line 255: The controller shouldn't have any kind of logic, you can do this on service side

rpvsilva (changes requested):

backend/src/modules/teams/controller/team.controller.ts
patricia-mdias: The service function delete() returns a boolean, indicating whether the delete query was successful or not, I believe I need this kind of information to return a Bad Request Status in the Controller. What do you think?

nunocaseiro: You can change the way we are handling the errors by throwing the exception on the service and just calling the service method in the controller

nunocaseiro (changes requested):

nunocaseiro (dismissed):
LGTM 🚀 great job!

rpvsilva (changes requested):

rpvsilva (approved):
LGTM 🔥

PR #652, opened 2022-12-02
feat: user repository

nunocaseiro:
## Proposed Changes

  - Add user repository
  - Change user._id to user.id


backend/src/libs/repositories/mongo/mongo-generic.repository.ts
```diff
@@ -0,0 +1,44 @@
...
+export class MongoGenericRepository<T> implements BaseAbstractRepository<T> {
+	private _repository: Model<T>;
+	private _populateOnFind: string[];
+
+	constructor(repository: Model<T>, populateOnFind: string[] = []) {
+		this._repository = repository;
+		this._populateOnFind = populateOnFind;
+	}
+
+	getAll(selectedValues?: string): Promise<T[]> {
```
rpvsilva, line 13: Is it possible that this `selectedValues`be an object where the keys are the model properties?

rpvsilva (changes requested):

backend/src/libs/repositories/mongo/mongo-generic.repository.ts
rpvsilva: You can talk with me to do this 😉

PR #662, opened 2022-12-06
feat: automatically update isnewjoiner status

### LINKED ISSUE #636: [FEATURE]: automatically update isNewJoiner status
## Summary

Currently, every-time that a member is added to a team, the attribute isNewJoiner is by default false, but a team admin or superAdmin can change that status in a swipe button. However, this situation is not the most efficient as an admin has to update by hand whenever a new member joins the company. 

Therefore, it is necessary to create a mechanism that allows updating the isNewJoiner attribute automatically, whenever a new element enters the company. In addition, it is necessary to create a mechanism that, after 3 months, resets the newJoiners status to false.

To ensure that only new members have isNewJoiner set to true, it should be possible to access the date when the azure email account was created, and check if it is included in the trial period (current date minus 3 months).

Whenever a retro is created, it should be checked whether the members of that team are still new joiners or not. If they have been with the company for more than 3 months, then they are no longer new joiners.

CatiaAntunes96:
Relates to #636 

## Proposed Changes

  - Change the logic of creating the teams of the subboards.
  - If there aren't enough members who can be responsible for a subboard, the oldest new joiner user will be chosen
  - Everytime a board is created, a verification function is executed, to check if the new joiners are still new joiners, updating accordingly.

This pull request closes #636

### DIFF
--- a/backend/src/modules/auth/shared/login.auth.ts
+++ b/backend/src/modules/auth/shared/login.auth.ts
@@ -8,10 +8,10 @@ export const signIn = async (
 	getTokenService: GetTokenAuthService | GetTokenAuthApplication,
 	strategy: string
 ) => {
-	const { email, firstName, lastName, _id, isSAdmin } = user;
+	const { email, firstName, lastName, _id, isSAdmin, userAzureCreatedAt } = user;
 	const jwt = await getTokenService.getTokens(_id);
 
 	if (!jwt) return null;
 
-	return { ...jwt, email, firstName, lastName, strategy, _id, isSAdmin };
+	return { ...jwt, email, firstName, lastName, strategy, id: _id, isSAdmin, userAzureCreatedAt };
 };
--- a/backend/src/modules/auth/shared/login.auth.ts
+++ b/backend/src/modules/auth/shared/login.auth.ts
@@ -13,5 +13,15 @@ export const signIn = async (
 
 	if (!jwt) return null;
 
-	return { ...jwt, email, firstName, lastName, strategy, id: _id, isSAdmin, userAzureCreatedAt };
+	return {
+		...jwt,
+		email,
+		firstName,
+		lastName,
+		strategy,
+		id: _id,
+		isSAdmin,
+		userAzureCreatedAt,
+		_id
+	};
 };
--- a/backend/src/modules/auth/shared/login.auth.ts
+++ b/backend/src/modules/auth/shared/login.auth.ts
@@ -8,7 +8,7 @@ export const signIn = async (
 	getTokenService: GetTokenAuthService | GetTokenAuthApplication,
 	strategy: string
 ) => {
-	const { email, firstName, lastName, _id, isSAdmin, userAzureCreatedAt } = user;
+	const { email, firstName, lastName, _id, isSAdmin, providerAccountCreatedAt } = user;
 	const jwt = await getTokenService.getTokens(_id);
 
 	if (!jwt) return null;
@@ -21,7 +21,7 @@ export const signIn = async (
 		strategy,
 		id: _id,
 		isSAdmin,
-		userAzureCreatedAt,
+		providerAccountCreatedAt,
 		_id
 	};
 };
--- a/backend/src/modules/azure/services/auth.azure.service.ts
+++ b/backend/src/modules/azure/services/auth.azure.service.ts
@@ -63,7 +63,7 @@ export default class AuthAzureServiceImpl implements AuthAzureService {
 			email: emailOrUniqueName,
 			firstName,
 			lastName,
-			userAzureCreatedAt: userFromAzure.value[0].createdDateTime
+			providerAccountCreatedAt: userFromAzure.value[0].createdDateTime
 		});
 
 		if (!createdUser) return null;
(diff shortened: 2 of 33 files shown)


backend/src/modules/auth/shared/login.auth.ts
```diff
@@ -8,10 +8,20 @@ export const signIn = async (
...
-	return { ...jwt, email, firstName, lastName, strategy, _id, isSAdmin };
+	return {
+		...jwt,
+		email,
+		firstName,
+		lastName,
+		strategy,
+		id: _id,
+		isSAdmin,
+		userAzureCreatedAt,
```
rpvsilva, line 24: Since this is a shared logic within the authentication, is it a good idea to call this property `userAzureCreatedAt`, imagine that I login using GitHub? What do you think we could do here? @nunocaseiro

backend/src/modules/azure/services/auth.azure.service.ts
```diff
@@ -70,10 +72,10 @@ export default class AuthAzureServiceImpl implements AuthAzureService {
 	}
 
 	getGraphQueryUrl(email: string) {
-		return `https://graph.microsoft.com/v1.0/users?$search="mail:${email}" OR "displayName:${email}" OR "userPrincipalName:${email}"&$orderbydisplayName&$count=true`;
+		return `https://graph.microsoft.com/v1.0/users?$search="mail:${email}" OR "displayName:${email}" OR "userPrincipalName:${email}"&$orderbydisplayName&$select=displayName,mail,userPrincipalName,createdDateTime`;
```
rpvsilva, line 75: Could you extract this url to a separate function outside the class? You can use it like:
```javascript
const Endpoints = {
  getAzureUserInfo: (email: string) => `...`
}
```
And then use it like `Endpoints.getAzureUserInfo(email)`. What do you think?

backend/src/modules/azure/services/auth.azure.service.ts
```diff
@@ -70,10 +72,10 @@ export default class AuthAzureServiceImpl implements AuthAzureService {
...
 	}
 
-	async checkUserExistsInActiveDirectory(email: string) {
+	async getUserFromAzure(email: string) {
```
rpvsilva, line 78: Related to above, then you just need this method. Also we need to remove this "normal" axios and use the axios module from nestjs, but that's different topic

backend/src/modules/boards/services/create.board.service.ts
```diff
@@ -165,14 +169,46 @@ export default class CreateBoardServiceImpl implements CreateBoardService {
...
 	}
 
+	verifyIfIsNewJoiner = (userAzureCreatedAt: Date | undefined, joinedAt: Date) => {
+		const currentDate = dayjs();
+
+		const dateToCompare = userAzureCreatedAt ? dayjs(userAzureCreatedAt) : dayjs(joinedAt);
+
+		const maxDateToBeNewJoiner = dateToCompare.add(3, 'month');
+
+		return currentDate.isBefore(maxDateToBeNewJoiner) || currentDate.isSame(maxDateToBeNewJoiner);
```
rpvsilva, line 179: As a personal preference I suggest using `date-fns` because it uses the native `Date` class from JS instead of creating an instance like `moment`, can you check it?

rpvsilva (changes requested):

backend/src/modules/auth/shared/login.auth.ts
nunocaseiro: Yeah I think we should change the var name to something more generic

CatiaAntunes96: What do you think about  **providerAccountCreatedAt** for the name of the variable? @nunocaseiro @rpvsilva

backend/src/modules/azure/services/auth.azure.service.ts
CatiaAntunes96: It seems a good idea. What do you think about creating a new file , called endpoints, in the libs/utils folder, and then use it like you suggested? @rpvsilva

backend/src/modules/boards/services/create.board.service.ts
CatiaAntunes96: You are right, like moment, dayjs also returns an instance. I will use `date-fns`

backend/src/modules/azure/services/auth.azure.service.ts
rpvsilva: I don't think that this is necessary, keep this things in the domain

backend/src/modules/auth/shared/login.auth.ts
rpvsilva: Seems good to me 😉

frontend/package.json
```diff
@@ -35,6 +35,7 @@
 		"@testing-library/dom": "^8.13.0",
 		"@testing-library/react": "^12.1.5",
 		"axios": "^0.26.1",
+		"dayjs": "^1.11.7",
```
rpvsilva, line 38: Could you install de `date-fns` on frontend too? To keep a standard on both sides

rpvsilva (changes requested):

rpvsilva (changes requested):

rpvsilva (approved):
For me backend is good 🔥 wait until nuno's approval for frontend

nunocaseiro (dismissed):
Amazing job 🔥

rpvsilva (approved):

PR #667, opened 2022-12-12
feat: change team on create board

### LINKED ISSUE #577: [BUG] Cancel board create button as non-admin/no team
##Description
As a user without admin role/no team, I should be able to cancel board creation as I do not have privileges to create them

###Current behavior
The button is greyed out for users that aren't admin or have a team

Image

### LINKED ISSUE #580: [BUG] Limit votes floating point
## Description
As an admin I should be able to set the max voting limit and be warned if there's something wrong

### Current behavior
Using floats instead of integers keeps the input box green, this box should only allow integers higher than 0 (some browsers will warn about the error upon submit)

Image

### LINKED ISSUE #663: [FEATURE]: change team on create board
## Summary

Currently, when a new board is created, is not possible to choose to which team the board will belong (by default, it will belong to the first team returned by the getAllTeams).

Then, it should be possible to view a list of the teams that the user belongs and  choose to which one the board will be associated. To accomplish this, all the logic on the frontend should be changed and the design should be similar to the figma. Besides, whenever we change the selected team, it should be revalidated to assure that the team has enough people to form multiple teams and that the user that is creating the board is the admin of that team.

CatiaAntunes96:
Relates to #663 
Fixes #577 
Fixes #580 

image

## Proposed Changes

  - Add a select with teams that the user is part of. If the user is super Admin, he can view all the teams.
  - When the user selects a team, it's validated if he is a team admin, a stakeholder or super admin, and if the team has the minimum number of members to form subteams. If one of this conditions fails, a message error is displayed and the user can select another team.
  - The joi schema validates if a team is selected. This prevents the user of creating a board without a team, if he clicks on the create button. 
  - Fix: Remove error message when the page is refreshed.
  - Fix: Cancel board create button bug fixed. The button is always accessible to the user. 
  - Fix: Max Votes input only accepts integers

This pull request closes #663 #577 #580

### DIFF
--- a/frontend/src/components/CreateBoard/SubTeamsTab/TeamSubTeamsConfigurations.tsx
+++ b/frontend/src/components/CreateBoard/SubTeamsTab/TeamSubTeamsConfigurations.tsx
@@ -27,7 +27,7 @@ import QuickEditSubTeams from './QuickEditSubTeams';
 import SelectTeam from './SelectTeam';
 import FakeMainBoardCard from '../fake/FakeSettingsTabs/partials/MainBoardCard';
 
-const StyledBox = styled(Flex, Box, { borderRadius: '$12', backgroundColor: 'white' });
+const StyledBox = styled(Flex, Box, {});
 
 type TeamSubTeamsInterface = {
   timesOpen: number;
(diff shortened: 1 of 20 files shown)


frontend/src/components/CreateBoard/SubTeamsTab/TeamSubTeamsConfigurations.tsx
```diff
@@ -91,62 +85,76 @@ const TeamSubTeamsConfigurations: React.FC<TeamSubTeamsInterface> = ({
...
+              pr: '$16',
+              borderRadius: '$4',
+              border: '1px solid $primary200',
+              background: '$background',
+              height: '$64',
+            }}
+            direction="column"
+            elevation="1"
+            gap="2"
+          >
```
nunocaseiro, line 105: Can you extract this component to the top of the file as you are using it twice with the same css and props?

frontend/src/pages/boards/new.tsx
```diff
@@ -51,6 +54,8 @@ const NewBoard: NextPage = () => {
   const boardState = useRecoilValue(createBoardDataState);
   const resetBoardState = useResetRecoilState(createBoardDataState);
```
nunocaseiro, line 55:
```suggestion
  const [boardState, setBoardState] = useRecoilState(createBoardDataState);
```

nunocaseiro (changes requested):

nunocaseiro (approved):
Great job 🔥 🚀

Issue #594, opened 2022-11-18
[FEATURE]: User management


nunocaseiro:
## Summary

As a super admin, I want to manage the platform users
As a member, I want to see all the platform users

### Basic example

<img width="1440" alt="UM 1 0 - User Management" src="https://user-images.githubusercontent.com/90208434/202678446-9a5d72c2-c732-4549-8cb4-3785df8933cd.png">

### Motivation

We need this in order to manage the role of the platform users and be able to assign teams

PR #682, opened 2022-12-14
feat: users lazy loading backend

### LINKED ISSUE #602: [FEATURE]: Users list lazy loading strategy
## Summary

A lazy loading strategy should be implemented to avoid fetching all the platform users at the same time. So the user can fetch, for example, 10 at a time and more 10 when the scroll reaches its end.
The endpoint created on issue #597 should be modified to receive some pagination parameters (offset and page)

## User story it relates to

#594

RafaelSBatista97:
#594 

Relates to #602 

image

image

## Proposed Changes

-A lazy loading strategy should be implemented to avoid fetching all the platform users at the same time. So the user can fetch, for example, 15 at a time and more 15 when the scroll reaches its end.

@GuiSanto 

This pull request closes #602

### DIFF
--- a/backend/src/modules/teams/repositories/team-user.repository.ts
+++ b/backend/src/modules/teams/repositories/team-user.repository.ts
@@ -1,7 +1,9 @@
 import { Injectable } from '@nestjs/common';
 import { InjectModel } from '@nestjs/mongoose';
+import { ObjectId } from 'mongodb';
 import { Model } from 'mongoose';
 import { MongoGenericRepository } from 'src/libs/repositories/mongo/mongo-generic.repository';
+import User from 'src/modules/users/entities/user.schema';
 import { UserWithTeams } from 'src/modules/users/interfaces/type-user-with-teams';
 import TeamUserDto from '../dto/team.user.dto';
 import TeamUser, { TeamUserDocument } from '../entities/team.user.schema';
@@ -38,15 +40,21 @@ export class TeamUserRepository
 		});
 	}
 
-	getUsersOnlyWithTeams(): Promise<UserWithTeams[]> {
+	getUsersOnlyWithTeams(users: User[]): Promise<UserWithTeams[]> {
+		const ids = users.map((user) => new ObjectId(user._id));
+
 		return this._repository
 			.aggregate([
+				{
+					$match: { $expr: { $in: ['$user', ids] } }
+				},
 				{
 					$lookup: {
 						from: 'users',
 						localField: 'user',
 						foreignField: '_id',
-						as: 'user'
+						as: 'user',
+						pipeline: [{ $sort: { firstName: 1 } }]
 					}
 				},
 				{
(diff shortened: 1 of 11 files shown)


backend/src/modules/teams/repositories/team-user.repository.ts
```diff
@@ -38,15 +40,21 @@ export class TeamUserRepository
...
+					$match: { $expr: { $in: ['$user', ids] } }
+				},
 				{
 					$lookup: {
 						from: 'users',
 						localField: 'user',
 						foreignField: '_id',
-						as: 'user'
+						as: 'user',
+						pipeline: [{ $sort: { firstName: 1 } }]
```
nunocaseiro, line 57:
```suggestion
```

backend/src/modules/teams/repositories/team-user.repository.ts
```diff
@@ -73,6 +80,7 @@ export class TeamUserRepository
 				{
 					$set: { userWithTeam: '$_id' }
 				},
+
```
nunocaseiro, line 83:
```suggestion
```

backend/src/modules/users/repository/user.repository.ts
```diff
@@ -36,4 +36,13 @@ export class UserRepository
 	updateSuperAdmin(userId: string, isSAdmin: boolean) {
 		return this.findOneByFieldAndUpdate({ _id: userId }, { $set: { isSAdmin } }, { new: true });
 	}
+
+	getAllWithPagination(page: number, size: number) {
+		return this.model
+			.find()
+			.skip(page * size)
+			.limit(size)
+			.sort({ firstName: 1 })
```
nunocaseiro, line 45: You are sorting here only by the first name but in the other method you are sorting by first or last name. Please check if this won't be a problem :)

nunocaseiro (changes requested):

backend/src/modules/users/services/get.user.service.ts
```diff
@@ -42,20 +42,43 @@ export default class GetUserServiceImpl implements GetUserService {
...
+			hasNextPage,
+			page
+		};
+
+		results.userWithTeams.sort((a, b) => {
+			if (a.user.firstName === b.user.firstName) {
+				return a.user.lastName < b.user.lastName ? -1 : 1;
+			} else {
+				return a.user.firstName < b.user.firstName ? -1 : 1;
+			}
```
rpvsilva, line 79:
```suggestion
			}
			return a.user.firstName < b.user.firstName ? -1 : 1;
```

rpvsilva (changes requested):

backend/src/modules/users/repository/user.repository.ts
RafaelSBatista97: maybe this can be removed because we are sorting every user on the other method?

nunocaseiro: The sort should also be here but make sure this Will work in every case

rpvsilva (approved):

nunocaseiro (approved):
LGTM 🚀

PR #711, opened 2022-12-19
fix: voting updates

### LINKED ISSUE #710: [BUG]: votes in card items. performance
## Description

The voting request in the card items doesn't return the whole board document so the frontend throws an error when trying to access the board users
Voting take some time to handle the response from the backend

nunocaseiro:
Fixes #710

## Proposed Changes

  - Return board users data on voting updates
  - Change the voting optimistic updates

This pull request closes #710


frontend/src/hooks/useVotes.tsx
```diff
@@ -65,6 +72,19 @@ const useVotes = () => {
...
+    boardData.users = boardData.users.map((boardUser) => {
+      if (boardUser.user._id === userId) {
+        return {
+          ...boardUser,
+          votesCount: action === Action.Add ? boardUser.votesCount + 1 : boardUser.votesCount - 1,
+        };
+      }
+
+      return boardUser;
+    });
```
rpvsilva, line 85: You could do the `if (boardUser.user._id !== userId) return boardUser;` first just to be cleaner

rpvsilva (changes requested):
Just a small suggestion, but LGTM

frontend/src/hooks/useVotes.tsx
nunocaseiro: of course :p

rpvsilva (approved):

Issue #750, opened 2022-12-26
[FEATURE]: Filter teams


GuiSanto:
## Summary

As a Super Admin, when I see the list of teams to add to a user, I should be able to type on the search bar, using throttle mechanism.

## User story it relates to

#594 

### Motivation

It becomes easier to find the team the users wants.

PR #759, opened 2022-12-29
feat: fetch teams, add teams to user, filter by search

### LINKED ISSUE #749: [FEATURE]: Add new teams to a user
## Summary

As a Super Admin, I should be able to click on the checkboxes next to the teams, and then when I click on the "Add" button, it closes the side pop-up and adds those new teams to the list.

## User story it relates to

#594 

### Basic example

image

image

### Motivation

Responsive layout.

GuiSanto:
Relates to #594

image

## Proposed Changes

  - fetch teams user is not member
  - check and uncheck teams
  - "Add" updates the DB, adding the selected teams to the user. Then it closes the dialog

This pull request closes #749 and #750

### DIFF
--- a/frontend/src/components/Users/UserEdit/partials/TeamsDialog/index.tsx
+++ b/frontend/src/components/Users/UserEdit/partials/TeamsDialog/index.tsx
@@ -1,14 +1,9 @@
-import React, { Dispatch, SetStateAction, useMemo, useRef, useState } from 'react';
-// import { useSession } from 'next-auth/react';
-import { useRecoilState } from 'recoil';
+import React, { Dispatch, SetStateAction, useEffect, useRef, useState, useMemo } from 'react';
 import { Dialog, DialogClose, DialogTrigger, Portal } from '@radix-ui/react-dialog';
 
 import Icon from '@/components/icons/Icon';
 import Text from '@/components/Primitives/Text';
-import { usersListState, userTeamsListState } from '@/store/team/atom/team.atom';
-// import { toastState } from '@/store/toast/atom/toast.atom';
-// import { TeamUserRoles } from '@/utils/enums/team.user.roles';
-// import { ToastStateEnum } from '@/utils/enums/toast-types';
+import { TeamUserRoles } from '@/utils/enums/team.user.roles';
 
 import {
   ButtonsContainer,
@@ -21,12 +16,13 @@ import {
 import Flex from '@/components/Primitives/Flex';
 import Checkbox from '@/components/Primitives/Checkbox';
 import Button from '@/components/Primitives/Button';
-// import { CreateTeamUser, TeamUserAddAndRemove } from '@/types/team/team.user';
-// import { useRouter } from 'next/router';
+import { useRouter } from 'next/router';
 
-// import { verifyIfIsNewJoiner } from '@/utils/verifyIfIsNewJoiner';
-// import useTeam from '@/hooks/useTeam';
+import { verifyIfIsNewJoiner } from '@/utils/verifyIfIsNewJoiner';
+import useTeam from '@/hooks/useTeam';
 import { AddNewBoardButton } from '@/components/layouts/DashboardLayout/styles';
+import { TeamChecked } from '@/types/team/team';
+import { TeamUserUpdate } from '@/types/team/team.user';
 import SearchInput from './SearchInput';
 import { ScrollableContent } from './styles';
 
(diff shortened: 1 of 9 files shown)


frontend/src/components/Users/UserEdit/partials/TeamsDialog/index.tsx
```diff
@@ -36,17 +32,40 @@ type Props = {
...
+    if (didMount.current) {
+      refetch();
+    }
+    didMount.current = true;
+  }, [isOpen, refetch]);
+
+  // after fetching data, add the field "isChecked", to be used in the Add button
+  if (data) {
+    teamsUserIsNotMember = data?.map((team) => ({ ...team, isChecked: false }));
+  }
```
rpvsilva, line 67:
```suggestion
  teamsUserIsNotMember = data?.map((team) => ({ ...team, isChecked: false })) || [];
```

frontend/src/components/Users/UserEdit/partials/TeamsDialog/index.tsx
```diff
@@ -57,14 +76,38 @@ const ListTeams = ({ isOpen, setIsOpen }: Props) => {
...
+        role: TeamUserRoles.MEMBER,
+        team: team._id,
+        isNewJoiner: verifyIfIsNewJoiner(
+          joinedAt as string,
+          providerAccountCreatedAt ? (providerAccountCreatedAt as string) : undefined,
+        ),
+      }));
+      mutate(teamUsers);
+      setIsOpen(false);
+    }
```
rpvsilva, line 101: You can use the flatMap to get the filtered teams by the `isChecked` property and transform the object.

frontend/src/components/Users/UsersList/partials/CardUser/EditUser.tsx
```diff
@@ -8,7 +8,16 @@ type EditUserProps = { user: User };
...
+    <Link
+      href={ROUTES.UserEdit(
+        user._id,
+        user.firstName,
+        user.lastName,
+        user.isSAdmin,
+        user.joinedAt,
+        user.providerAccountCreatedAt,
+      )}
+    >
```
rpvsilva, line 20: You could destructure the user object

frontend/src/hooks/useTeam.tsx
```diff
@@ -125,6 +145,25 @@ const useTeam = ({
     },
   });
 
+  const updateAddTeamsToUser = useMutation(updateAddTeamsToUserRequest, {
+    onSuccess: async () => {
+      await Promise.all([queryClient.invalidateQueries(['teams', userId])]);
```
rpvsilva, line 150: You don't need the `await Promise.all` since you are awaiting for just one promise

frontend/src/utils/routes.ts
```diff
@@ -18,8 +18,15 @@ export const ROUTES = {
...
-  UserEdit: (userId: string, firstName: string, lastName: string, isSAdmin: boolean) =>
-    `/users/${userId}?firstName=${firstName}&lastName=${lastName}&isSAdmin=${isSAdmin}`,
+  UserEdit: (
+    userId: string,
+    firstName: string,
+    lastName: string,
+    isSAdmin: boolean,
+    joinedAt: string,
+    providerAccountCreatedAt?: string,
+  ) =>
```
rpvsilva, line 28: Instead of receiving many parameters you could receive user parameter, destructure it and the use its properties. This way you can only pass the user object when using this endpoint, like in the above usage

rpvsilva (changes requested):

frontend/src/components/Users/UserEdit/partials/TeamsDialog/index.tsx
RafaelSBatista97: What do you mean? Didn't understand

rpvsilva: You are filtering the `teamsUserIsNotMember` by the `isChecked` property. then you do a `.map()` to map the fields. you can do all of this with a single `flatMap`

GuiSanto: image

Flatmap should remove the undefineds, but its not?

rpvsilva: Check the others PRs that you have used flatMap, remember what I said about the flatMap? Lowering a level?

GuiSanto: Yes but I don't know why we need to lower a level because we don't use the values of the lower levels

rpvsilva (approved):
LGTM 🔥

PR #764, opened 2022-12-30
fix: only makes request when the dialog is opened

### LINKED ISSUE #763: [BUG]: a request is made every time the "add new team" dialog is open/close
## Description

A network request is being made every time the dialog opens, on the Edit User page.

### Steps to reproduce

Click to open and close the dialog.

### Expected result

Only one request when the dialog is opened.

### Actual result

Requests when it both opens and closes

RafaelSBatista97:
#763 

Fixes #763

## Proposed Changes

  - Change the logic used on useEffect to get the data. 
  - Don't use useRef
 

@GuiSanto 

This pull request closes #763


rpvsilva (approved):
LGTM 🔥

PR #772, opened 2023-01-02
feat: participants tab select participants

### LINKED ISSUE #678: [FEATURE]: Participants Tab - Select Participants
## Summary

If the "Select Participants", a dialog with all the users similar to the one in the "Create Team" should be rendered to select the participants. The user that is creating the team should be selected.

## User story it relates to

#674 

### Basic example
image

CatiaAntunes96:
Relates to #678 

## Proposed Changes

  - When the user selects the option "Select Participants", a box with the user which is creating the team and a button are renderer.
  - When the user clicks on the button, it opens a dialog with all the users .

This pull request closes #678


nunocaseiro (approved):
LGTM 🚀

PR #785, opened 2023-01-04
feat: create regular board action backend

### LINKED ISSUE #680: [FEATURE]: create regular board action
## Summary

After the regular board has a title and a team/participants are selected, when the user clicks on the "Create Board" button, only a main board should be created and the responsible of the board will be the user that is creating the board. 
Also, the backend logic needs to be modified in order to accept both types of boards. For this, you can send along with the board that will be created, its type ("split", "retro"), which will allow the backend what kind of approach to follow. Also, it is necessary to add logic to create a regular board.

## User story it relates to

#674

CatiaAntunes96:
Relates to #680 

## Proposed Changes

  - Adapt create function to create a regular board
  - Fix delete board to also delete a regular board with only participants
  - Fix getBoards function to also retrieve regular boards (all the boards from an user)

This pull request closes #680

### DIFF
--- a/backend/src/modules/boards/controller/boards.controller.ts
+++ b/backend/src/modules/boards/controller/boards.controller.ts
@@ -239,11 +239,7 @@ export default class BoardsController {
 		@Query() { socketId }: BaseParamWSocket,
 		@Req() request: RequestWithUser
 	) {
-		const result = await this.deleteBoardApp.delete(
-			boardId,
-			request.user._id,
-			request.user.isSAdmin
-		);
+		const result = await this.deleteBoardApp.delete(boardId, request.user);
 
 		if (socketId && teamId) {
 			this.socketService.sendUpdatedBoards(socketId, teamId);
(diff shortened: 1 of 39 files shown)


nunocaseiro (dismissed):
LGTM 🚀

backend/src/modules/boards/controller/boards.controller.ts
```diff
@@ -239,7 +239,11 @@ export default class BoardsController {
 		@Query() { socketId }: BaseParamWSocket,
 		@Req() request: RequestWithUser
 	) {
-		const result = await this.deleteBoardApp.delete(boardId, request.user._id);
+		const result = await this.deleteBoardApp.delete(
+			boardId,
+			request.user._id,
+			request.user.isSAdmin
+		);
```
rpvsilva, line 246: Can you refactor this method and move this logic to the service side?

backend/src/modules/boards/services/get.board.service.ts
```diff
@@ -79,7 +79,7 @@ export default class GetBoardServiceImpl implements GetBoardServiceInterface {
 		return this.getBoards(false, query, page, size);
 	}
 
-	getTeamBoards(teamId: string, page: number, size?: number) {
+	async getTeamBoards(teamId: string, page: number, size?: number) {
```
rpvsilva, line 82:
```suggestion
	getTeamBoards(teamId: string, page: number, size?: number) {
```

rpvsilva (changes requested):

nunocaseiro (approved):

PR #786, opened 2023-01-04
feat: create regular board action frontend

### LINKED ISSUE #680: [FEATURE]: create regular board action

CatiaAntunes96:
Relates to #680 

## Proposed Changes

  - Add saveBoard function to create a regular board
  - Fix socket erros when the user has regular boards with only participants
  - The boards list is receiving all the boards, but for now the it only shows boards with teams because of the function that is being called on the filter. 

This pull request closes #680

### DIFF
--- a/frontend/src/pages/boards/newRegularBoard.tsx
+++ b/frontend/src/pages/boards/newRegularBoard.tsx
@@ -37,6 +37,7 @@ import { TeamUserRoles } from '@/utils/enums/team.user.roles';
 import SchemaCreateRegularBoard from '@/schema/schemaCreateRegularBoard';
 import { getAllUsers } from '@/api/userService';
 import { ToastStateEnum } from '@/utils/enums/toast-types';
+import useBoard from '@/hooks/useBoard';
 
 const defaultBoard = {
   users: [],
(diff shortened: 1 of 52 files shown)


frontend/src/pages/boards/newRegularBoard.tsx
```diff
@@ -76,8 +77,8 @@ const NewRegularBoard: NextPage = () => {
   const [createBoard, setCreateBoard] = useState(false);
 
   const setToastState = useSetRecoilState(toastState);
-  const setBoardState = useSetRecoilState(createBoardDataState);
-  // const [boardState, setBoardState] = useRecoilState(createBoardDataState);
+  // const setBoardState = useSetRecoilState(createBoardDataState);
```
nunocaseiro, line 80:
```suggestion
```

nunocaseiro (approved):
LGTM 🚀

nunocaseiro (approved):

PR #787, opened 2023-01-04
fix: votes

### LINKED ISSUE #778: [BUG] Board votes inconsistent
## Description
As a user I should be able to vote on the main board

### Current behavior
It's possible for a user to bypass the main board vote limit, this should be looked into as it's possible for a user to end up with way more votes than they should have

Limited votes to 10, but user with the ID 638f5968ca851b9a64a4ea8b has 21 votes on a single card
Image

nunocaseiro:
Relates to #778

## Proposed Changes

  - Add debounce to add or remove votes
  - The votes are now counted and after 300ms the request is sent with the count value
  - The backend now uses the count that is sent from the frontend to decrease or increase votes

This pull request closes #778

### DIFF
--- a/backend/src/modules/votes/controller/votes.controller.ts
+++ b/backend/src/modules/votes/controller/votes.controller.ts
@@ -1,22 +1,10 @@
-import {
-	BadRequestException,
-	Body,
-	Controller,
-	Delete,
-	Inject,
-	Param,
-	Post,
-	Put,
-	Req,
-	UseGuards
-} from '@nestjs/common';
+import { Body, Controller, HttpStatus, Inject, Param, Put, Req, UseGuards } from '@nestjs/common';
 import {
 	ApiBadRequestResponse,
 	ApiBearerAuth,
 	ApiBody,
 	ApiCreatedResponse,
 	ApiInternalServerErrorResponse,
-	ApiOkResponse,
 	ApiOperation,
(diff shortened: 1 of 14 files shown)


backend/src/modules/votes/controller/votes.controller.ts
```diff
@@ -77,22 +63,27 @@ export default class VotesController {
...
 	) {
 		const { boardId, cardId, itemId } = params;
-		const board = await this.createVoteApp.addVoteToCard(boardId, cardId, request.user._id, itemId);
+		const { count, socketId } = data;
+
+		if (count < 0) {
+			await this.deleteVoteApp.deleteVoteFromCard(boardId, cardId, request.user._id, itemId, count);
+		} else {
+			await this.createVoteApp.addVoteToCard(boardId, cardId, request.user._id, itemId, count);
+		}
```
rpvsilva, line 79: This kind of logic must be on service side and not on controller

rpvsilva (changes requested):

backend/src/modules/votes/controller/votes.controller.ts
nunocaseiro: I understand but I have two services and the controller is the only handler that can decide which service should be called

rpvsilva (approved):

PR #822, opened 2023-01-06
feat: add archive service

### LINKED ISSUE #773: [FEATURE]: created channels for past retros should be archived
## Summary
All Slack channels should be removed when they are not necessary anymore, i.e. the previous channels should be removed.
Since removing them directly is impossible (in our scope), we will send all previously created channels to the Archived. Slack admin will handle it accordingly.

As a first approach, this will be done in recursive  Boards every time the cron job runs to create a new one it will archive the oldest one.

### Motivation
It will avoid polluting the Slack Workspace with unuseful channels.

mourabraz:
Relates to #773 

## Proposed Changes

  - add a new service inside the Communication module: ArchiveChannelService for SlackArchiveChannelApplication.
 
 The implementation for this service accepts a slack channel id or an object, as a second parameter it accepts a boolean (default value is false), in order to archive, or not, all channels related to the board (the property "dividedBoards" should contain the subboards).

This pull request closes #773


nunocaseiro:
Is this pr ready to review? I'm asking this because it is in draft mode but you request a review

backend/src/modules/communication/applications/slack-archive-channel.application.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+		application = new SlackArchiveChannelApplication(
+			new ConversationsSlackHandler(communicationGateAdapterMocked)
+		);
+	});
+
+	it('should be defined', () => {
+		expect(application).toBeDefined();
+	});
+
+	it('shoult archive channel by channel id', async () => {
```
rpvsilva, line 75:
```suggestion
	it('should archive channel by channel id', async () => {
```

rpvsilva: Typo in every `it`

backend/src/modules/communication/applications/slack-archive-channel.application.ts
```diff
@@ -0,0 +1,81 @@
...
+			return [await this.archiveChannel(data.data)];
+		}
+
+		const board = data.data as BoardType;
+
+		if (!data.cascade) {
+			return [await this.archiveChannel(board.slackChannelId)];
+		}
+
+		return await this.arquiveAllChannelsInMainBoard(board);
```
rpvsilva, line 29:
```suggestion
		return this.arquiveAllChannelsInMainBoard(board);
```

backend/src/modules/communication/consumers/slack-archive-channel.consumer.ts
```diff
@@ -0,0 +1,48 @@
...
+			this.logger.verbose(
+				`Archive related channel(s) to board with id: "${
+					(data.data as BoardType).id
+				}" and Job id: "${job.id}"`
+			);
+		}
+
+		const result = await this.application.execute(data);
+
+		return result;
```
rpvsilva, line 46:
```suggestion
		return this.application.execute(data);
```

rpvsilva (dismissed):

backend/src/modules/communication/applications/slack-archive-channel.application.spec.ts
mourabraz: Done!

backend/src/modules/communication/applications/slack-archive-channel.application.ts
mourabraz: Done!

backend/src/modules/communication/applications/slack-archive-channel.application.ts
```diff
@@ -0,0 +1,81 @@
...
+	private async archiveChannel(channelId): Promise<ArchiveChannelResult> {
+		return { channelId, result: await this.conversationsHandler.archiveChannel(channelId) };
+	}
+
+	private async arquiveAllChannelsInMainBoard(board: BoardType): Promise<ArchiveChannelResult[]> {
+		const boards = [board, ...board.dividedBoards].filter(
+			(i) => typeof i.slackChannelId === 'string'
+		);
+
+		const promises = boards.map((i) => this.conversationsHandler.archiveChannel(i.slackChannelId));
```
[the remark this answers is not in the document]
mourabraz: I am not seeing the `flatMap` here. But I think keeping this two parts is more readable anyway... There are two different things happening, first we are filtering what we want then creating the promises, the predicate in the filter is not related to the promises.

Hum so, what do think about this way?

```js
const allSlackChannelIds = [board, ...board.dividedBoards]
			.filter((i) => typeof i.slackChannelId === 'string')
			.map((i) => i.slackChannelId);
		const promises = allSlackChannelIds.map((i) => this.conversationsHandler.archiveChannel(i));

		const results = await this.resolvePromises(promises);

		return allSlackChannelIds.map((channelId, idx) => ({
			channelId,
			result: results[idx]
		}));
```

backend/src/modules/communication/consumers/slack-archive-channel.consumer.ts
mourabraz: Done!

nunocaseiro (dismissed):

nunocaseiro (approved):

PR #827, opened 2023-01-07
fix: update card position and votes with sockets

### LINKED ISSUE #825: [BUG]: board performance issues
## Description

When are multiple users updating and voting on the cards the application the usage of cpu and memory increases a lot and it becomes unusable. The alternative is implementing sockets to avoid bottlenecks

nunocaseiro:
Fixes #825 

## Proposed Changes

Note: this should solve the majority of issues regarding performance when many people are moving or voting the cards in real-time. This should be tested in the live version

  - When voting in a card or moving its position now a socket message is sent to all users that are seeing the board and receiving that they will update the board locally instead of making a board request 
  - If some error occurs in the backend during these updates a message is sent to all users to refetch the board. (this may not be necessary)

This pull request closes #825

### DIFF
--- a/backend/src/modules/cards/dto/update-position.card..dto.ts
+++ b/backend/src/modules/cards/dto/update-position.card..dto.ts
@@ -1,4 +1,4 @@
-import { IsMongoId, IsNotEmpty, IsNumber, IsString } from 'class-validator';
+import { IsBoolean, IsMongoId, IsNotEmpty, IsNumber, IsString } from 'class-validator';
 import { BaseDto } from 'src/libs/dto/base.dto';
 
 export class UpdateCardPositionDto extends BaseDto {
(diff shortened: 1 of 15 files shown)


backend/src/modules/cards/dto/update-position.card..dto.ts
```diff
@@ -10,4 +10,31 @@ export class UpdateCardPositionDto extends BaseDto {
 	@IsNotEmpty()
 	@IsNumber()
 	newPosition!: number;
+
```
rpvsilva, line 13: The file is called `update-position.card..dto.ts` with double dots? 😂

frontend/src/hooks/useVotes.tsx
```diff
@@ -236,15 +236,14 @@ const useVotes = () => {
       variables,
     );
 
-    if (newBoardData?.maxVotes && newBoardData) {
+    if (newBoardData?.maxVotes && newBoardData && variables.userId === userId) {
```
rpvsilva, line 239:
```suggestion
    if (newBoardData?.maxVotes && variables.userId === userId) {
```

rpvsilva (changes requested):

backend/src/modules/cards/dto/update-position.card..dto.ts
nunocaseiro: ahah nice catch

rpvsilva (approved):

PR #828, opened 2023-01-08
fix: validation text area on change

nunocaseiro:
## Proposed Changes

  - Remove on change in the text area. It should be handled by the react-hook-form


backend/src/modules/cards/dto/update-position.card.dto.ts
```diff
@@ -1,13 +1,40 @@
-import { IsMongoId, IsNotEmpty, IsNumber, IsString } from 'class-validator';
+import { IsBoolean, IsMongoId, IsNotEmpty, IsNumber, IsString } from 'class-validator';
 import { BaseDto } from 'src/libs/dto/base.dto';
 
 export class UpdateCardPositionDto extends BaseDto {
 	@IsNotEmpty()
-	@IsString()
 	@IsMongoId()
+	@IsString()
```
rpvsilva, line 7: Since you have `@IsMongoId` you don't need the isString validation

rpvsilva (changes requested):

rpvsilva (approved):

PR #853, opened 2023-01-11
fix: board loading

### LINKED ISSUE #846: [BUG] Board loading
## Description
As a user I should be able to enter a board that I'm included

### Current behavior
As a super admin, interacting with a board that I'm not in, results in a loading (changes are still going through)
As a super admin, interacting with a board that I'm not in, results in a subboard page without main board.
As a normal user, interacting or entering a board that I'm not in, results in a loading (changes do not go through)

### Proposed changes
As a super admin, I should be able to interact with any board, even if I'm not in them
As a super admin, interacting with a board that I'm not in, the main board should always be available.
As a normal user, I should be kicked back to the dashboard the moment I try to enter a board that I'm not in (disallow any changes and/or entry)

Image

CatiaAntunes96:
Fixes #846 

## Proposed Changes

  - Fix superAdmin/ stakeholder access to the boards
  - Fix breadCrumbs when an user interacts with a board
  - Fix board count on the teams page
  - Remove mainBoardData object from response when a board is fetched

This pull request closes #846

### DIFF
--- a/backend/src/modules/boards/services/get.board.service.ts
+++ b/backend/src/modules/boards/services/get.board.service.ts
@@ -208,13 +208,13 @@ export default class GetBoardServiceImpl implements GetBoardServiceInterface {
 		// board1 = this.commentsClean(b)
 		board = this.cleanBoard(board, userId);
 
-		if (board.isSubBoard) {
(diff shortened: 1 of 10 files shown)


nunocaseiro (approved):
LGTM 🚀

PR #893, opened 2023-01-18
fix: bug login page design and bugs

### LINKED ISSUE #890: [BUG]: login page design and bugs
## Description

 - When logging in with invalid credentials we get a non-descriptive error message. And the position is also wrong
 - The Image has no Shadow (as seen in the design document)
 - The Password Input has a black icon instead of dark gray
image

### Expected result

image

JoaoSaIvador:
Relates to #890 
Fixes #890

image

## Proposed Changes

- When logging in with invalid credentials we get the proper error message
- The Image has a Shadow
- The Password Input a dark gray icon

This pull request closes #890

### DIFF
--- a/frontend/src/components/Primitives/Input.tsx
+++ b/frontend/src/components/Primitives/Input.tsx
@@ -235,11 +235,10 @@ const Input: React.FC<InputProps> = ({
   const isValueEmpty = isEmpty(value);
 
   const autoState = useMemo(() => {
-    if (Object.keys(errors).length > 0 || message) return 'error';
-    if (isValueEmpty) return 'default';
+    if (errors[id]) return 'error';
     if (!message && !isValueEmpty) return 'valid';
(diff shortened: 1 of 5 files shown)


frontend/src/components/Primitives/Input.tsx
```diff
@@ -235,7 +235,7 @@ const Input: React.FC<InputProps> = ({
   const isValueEmpty = isEmpty(value);
 
   const autoState = useMemo(() => {
-    if (message) return 'error';
+    if (Object.keys(errors).length > 0 || message) return 'error';
```
mourabraz, line 238: Since `message` value depends on the `error`, it will be truthy only when `Object.keys(errors).length > 0` is truthy as well, considering own properties of the object, in that case it can be removed.

nunocaseiro (approved):

PR #935, opened 2023-01-25
feat: storybook

### LINKED ISSUE #931: [FEATURE]: Storybook
image

## Summary

The web’s universality is pushing more complexity into the frontend. It began with responsive web design, which turned every user interface from one to 10, 100, 1000 different user interfaces. Over time, additional requirements piled on like devices, browsers, accessibility, performance, and async states.

Component-driven tools like Next.js help break down complex UIs into simple components but they’re not silver bullets. As frontends grow, the number of components swells. Mature projects can contain hundreds of components that yield thousands of discrete variations.

## Motivation

Every piece of UI is now a component. The superpower of components is that you don't need to spin up the whole app just to see how they render. You can render a specific variation in isolation by passing in props, mocking data, or faking events.

Storybook is packaged as a small, development-only, workshop that lives alongside your app. It provides an isolated iframe to render components without interference from app business logic and context. That helps you focus development on each variation of a component, even the hard-to-reach edge cases.

StereoPT:
Relates to #931
Fixes #931

image

## Proposed Changes

  - Added Storybook to SPLIT Project
  - Added simple Primitives to understand Storybook

This pull request closes #931


JoaoSaIvador (approved):
LGTM 🔥

PR #948, opened 2023-01-27
fix: set check value only on slackEnable

### LINKED ISSUE #934: [BUG] Anonymous checkmark
## Description
As a user I should be able to check the Post anonymously checkbox to make my messages anonymous

### Current behavior
The checkbox is fully black when checked, it should have a checkmark

Image

## Supposed behavior

Image

JoaoSaIvador:
Relates to #934 
Fixes #934 

## Screenshots

image

## Proposed Changes

  - Set check value only on slackEnable (so it is persistent when changing tabs when creating a board but doesn't affect the behaviour of the checkbox on other components)

This pull request closes #934

### DIFF
--- a/frontend/src/components/Primitives/Checkbox.tsx
+++ b/frontend/src/components/Primitives/Checkbox.tsx
@@ -100,7 +100,7 @@ const Checkbox: React.FC<{
   );
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


StereoPT (approved):
LGTM 🗝

Issue #958, opened 2023-01-30
[FEATURE]: delete column/cards from a regular board


CatiaAntunes96:
## Summary

As an user, when I'm on a regular board, I want to be able to remove all the cards from a column or delete the column. Those options are available on the menu settings of each column. To delete a board, we need to update the board without that column, but to delete all the cards from a column, we need to ensure that only this field is updated. Also, the votes have to be redistributed by all the users.

PR #973, opened 2023-01-31
feat: delete board column

patricia-mdias:
Relates to #958

## Proposed Changes

  - Fixed problem with votes count of a board user.
  - Deleting columns returns votes from the deleted cards.

@CatiaAntunes96

### DIFF
--- a/backend/src/modules/boards/services/update.board.service.ts
+++ b/backend/src/modules/boards/services/update.board.service.ts
@@ -27,6 +27,7 @@ import { UpdateColumnDto } from '../dto/column/update-column.dto';
 import { UPDATE_FAILED } from 'src/libs/exceptions/messages';
 import SocketGateway from 'src/modules/socket/gateway/socket.gateway';
 import { DeleteVoteServiceInterface } from 'src/modules/votes/interfaces/services/delete.vote.service.interface';
+import Column from '../schemas/column.schema';
 
 @Injectable()
 export default class UpdateBoardServiceImpl implements UpdateBoardServiceInterface {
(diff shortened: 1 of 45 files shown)


backend/src/modules/boards/services/update.board.service.ts
```diff
@@ -84,6 +88,16 @@ export default class UpdateBoardServiceImpl implements UpdateBoardServiceInterfa
...
+		if (boardData.deletedColumns && !isEmpty(boardData.deletedColumns)) {
+			const cardsToDelete = boardData.deletedColumns.flatMap((deletedColumnId: string) => {
+				return board.columns.find((column) => column._id.toString() === deletedColumnId)?.cards;
+			});
+
+			cardsToDelete.forEach(async (card) => {
+				await this.deleteCardService.delete(board.id, card._id, card.createdBy.toString(), card);
+			});
+		}
+
```
nunocaseiro, line 100: this code won't run in a split board update because you are checking if the `deleteColumns` field isn't empty. right?

backend/src/modules/cards/services/delete.card.service.ts
```diff
@@ -75,11 +76,12 @@ export default class DeleteCardServiceImpl implements DeleteCardService {
 		}
 	}
 
-	async delete(boardId: string, cardId: string, userId: string) {
+	async delete(boardId: string, cardId: string, userId: string, card?: Card) {
```
nunocaseiro, line 79: Why are you setting a card as an argument? To avoid one extra request to the database?

backend/src/modules/cards/services/delete.card.service.ts
```diff
@@ -172,7 +178,7 @@ export default class DeleteCardServiceImpl implements DeleteCardService {
 					{
 						$pull: {
 							'columns.$.cards.$[card].items': {
-								_id: cardItemId,
+								_id: cardId,
```
nunocaseiro, line 181: why are you changing this?

nunocaseiro (changes requested):

backend/src/modules/boards/services/update.board.service.ts
CatiaAntunes96: This code will only be executed if the deletedColumns field exists and isn't empty. When we send data from a split board, this field is always undefined.

backend/src/modules/boards/services/update.board.service.ts
```diff
@@ -84,6 +88,16 @@ export default class UpdateBoardServiceImpl implements UpdateBoardServiceInterfa
...
 		const { isSubBoard } = board;
 
+		if (boardData.deletedColumns && !isEmpty(boardData.deletedColumns)) {
+			const cardsToDelete = boardData.deletedColumns.flatMap((deletedColumnId: string) => {
+				return board.columns.find((column) => column._id.toString() === deletedColumnId)?.cards;
+			});
+
+			cardsToDelete.forEach(async (card) => {
+				await this.deleteCardService.delete(board.id, card._id, card.createdBy.toString(), card);
+			});
```
nunocaseiro, line 98: I think you don't have to delete each card. You just need to count the votes that will be returned to the user. Deleting the column should automatically delete all the cards.
If you delete the card two operations are executed, the deletion itself and the user votes incrementation

nunocaseiro (changes requested):

PR #994, opened 2023-02-02
feat: delete board columns/handle votes

CatiaAntunes96:
Relates to #958 

## Proposed Changes

- Fixed problem with votes count of a board user.
- Deleting columns returns votes from the deleted cards.

@patricia-mdias


backend/src/modules/boards/services/update.board.service.ts
```diff
@@ -164,6 +170,64 @@ export default class UpdateBoardServiceImpl implements UpdateBoardServiceInterfa
...
+					);
+
+					if (columnToDelete) {
+						return [];
+					}
+				}
+
+				return [{ ...col }];
+			}) as Column[];
+		}
```
nunocaseiro, line 229: Can you confirm if this code isn't executed in a split board?

nunocaseiro (dismissed):

backend/src/modules/boards/services/update.board.service.ts
CatiaAntunes96: I can confirm that this code isn't executed in a split board

PR #1018, opened 2023-02-03
feat: delete column/cards on a board

### LINKED ISSUE #738: [FEATURE]: Column operations
## Summary

If the user has the privileges of an admin:

-   And the board is a regular one he can:
    - Edit the column name
    - Remove all the cards of a column
    - Activate the card's default text
    - Delete a column (it should have at least one column)
    - Select the colour for the cards

Note: This feature can be divided into sub-features

## User story it relates to

#737

### Basic example

image

### LINKED ISSUE #958: [FEATURE]: delete column/cards from a regular board

CatiaAntunes96:
Relates to #958 
Relates to #738 

## Proposed Changes

On a regular board, a user with permitions can:
  - Empty cards from column
  - Delete column
  - Console error fixed

This pull request closes #958 
This pull request closes #738

### DIFF
--- a/frontend/src/hooks/useColumn.tsx
+++ b/frontend/src/hooks/useColumn.tsx
@@ -3,6 +3,7 @@ import BoardType from '@/types/board/board';
 import ColumnType from '@/types/column';
 import { ToastStateEnum } from '@/utils/enums/toast-types';
 import { deleteCardsFromColumnRequest, updateColumnRequest } from '@/api/boardService';
+import { BoardUser } from '@/types/board/board.user';
 import useBoardUtils from './useBoardUtils';
 
 const useColumn = () => {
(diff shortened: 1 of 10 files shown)


frontend/src/hooks/useColumn.tsx
```diff
@@ -67,8 +67,23 @@ const useColumn = () => {
     },
   });
 
+  const deleteCardsFromColumn = useMutation(deleteCardsFromColumnRequest, {
+    onSuccess: (data) => {
+      queryClient.invalidateQueries(['board', { id: data._id }]);
```
nunocaseiro, line 72: Does the endpoint return the board updated? if yes instead of invalidating the query you can use the response to update it

nunocaseiro (changes requested):

nunocaseiro (approved):

Issue #740, opened 2022-12-23
[FEATURE]: Board timer


nunocaseiro:
## Summary

All members should see a timer at the top of the board page. 
Only the participants with privileges should be able to interact with the timer. 
So a user that is responsible in a sub-board, is a stakeholder or team admin in the main board and is a team admin/creator in a regular board can interact with the timer.
The timer should be in sync with all members so the use of sockets may be needed 

## User story it relates to

#739

PR #1020, opened 2023-02-03
feat: board timer

### LINKED ISSUE #739: [FEATURE]: Board Timer
## User story it relates to

As a user with privileges in a board, I should be able to set, start, stop and pause a count down timer. 

### Basic example

image

geomarb:
Relates to #740 & #739

### Board with timer
Screenshot 2023-02-03 at 18 19 40

### Timer Set [`hasAdminRole` only]

The "arrows" located on the top and the bottom of the timer are present only when the timer is stopped for users with `hasAdminRole=true`. It allows setting the minutes and seconds of the timer, this action will set the timer for all users connected to that board via socket.

Screenshot 2023-02-03 at 18 21 46

### Timer STATUS
There are 3 possible status: RUNNING, STOPPED and PAUSED. Each time the status is changed it will be synced across all users on the same board.

#### RUNNING

Screenshot 2023-02-03 at 17 38 10

When the "play" button is pressed it will change the status to RUNNING. It means the timer will start counting down the seconds and minutes and send RUNNING instructions to all connected users.
On the server side, it will add 3 callback functions in the timer: Update, Expire, and Clean Timer.

##### Update Callback

This function will send the updated state to all connected clients, with duration, time left, and status. The client (browser) listens to this event and syncs the timer. It happens every 15 seconds if no value is passed to `BOARD_TIMER_SYNC_INTERVAL` which is now set to 5 seconds.

##### Expire Callback

This will run once the timer is over. By now it sends a stop operation. Stopping the running timers.

##### Clean Timer Callback

This function will remove the timer after 1 hour by default. This value is a variable if needed can be easily changed. It was created to avoid keeping garbage in memory for longer.

#### PAUSED

Screenshot 2023-02-03 at 17 38 27

When the pause button is clicked it will set the timer to PAUSED status and broadcast the status across all users connected to that board. In the backend, it will stop the interval callback, which means no updated status will be sent each 5 or 15 seconds until starts again. In the front end, it will keep running the interval, but now it will just alternate the `variant` style each second to give the blink feeling.

#### STOPPED

Screenshot 2023-02-03 at 17 37 51

When the stop button is clicked it will sync it across all users with that board open. It will remove the interval callback in the frontend and in the backend. The responsible can then set the timer again and or start it again.

## Timer STATE update

Each time a new user enters in a board it will request the status to the server, and the server will send it back, in this way even if someone has refreshed the browser or joined later in that board he will see the same timer state as others.

Important: the timer state is not persisted, if the server is restarted all timer states will be lost, it will cause sync issues.

## Dev notes

I used events in the backend to handle the communication between the socket module and the board module. The main advantage is that it loses coupling as one module doesn't need to know about the other. It also gives more flexibility, since, if needed another module can listen to these events and do what it needs without any changes in the emitter module.

This pull request closes #739


frontend/src/components/Board/SplitBoard/Timer/styles.tsx
```diff
@@ -0,0 +1,90 @@
+import { styled } from '@stitches/react';
```
nunocaseiro, line 1:
```suggestion
import { styled } from '@/styles/stitches/stitches.config';
```

nunocaseiro (changes requested):

nunocaseiro:
Before merging this I think we should check if everything is working well when there are multiple users updating the board at the same time. 

Just to finish, can you add the board timer in the regular boards?

geomarb:
Agree.

Timer added to regular boards.

nunocaseiro (approved):
LGTM 🚀

PR #1027, opened 2023-02-06
fix: populates

### LINKED ISSUE #986: [BUG] Also hide emojis if cards are hidden
## Description
As a sub-team responsible/admin/stakeholder I should be able to hide the cards' content

### Current behavior
Emojis are not getting obfuscated

Image

### LINKED ISSUE #1014: [BUG]: Content of all boards is being exposed
## Description

At the moment, the main board of January has all cards content hidden and if we try to inspect the element, we see this "aaaa":  

Screenshot 2023-02-03 at 15 13 27  

But if we go to the Network tab and check the response request, we can see the content of all cards:  

Screenshot 2023-02-03 at 15 15 37

Another issue, the sub-teams that are not merged to mainboard, it's possible to see their content cards. The example bellow is from team 3:

Screenshot 2023-02-03 at 15 16 53

### Steps to reproduce

1 - Open the network tab and refresh the main page
2 - Check the response of GET https://split-be.azurewebsites.net/boards/63da7ea1f1ffb3f94d1ad30f
Screenshot 2023-02-03 at 15 24 06

3 - Check the board.dividedBoards variable and you will see all the information
Screenshot 2023-02-03 at 15 24 35

### Expected result

What should happen?
This information should not be exposed

### LINKED ISSUE #1024: [BUG]: The blur causes a big performance issue
## Description

Using the safari, when a user visits a board page and it has the hide cards option enabled, the page becomes unresponsive and very laggy

nunocaseiro:
Fixes #986
Fixes #1024 
Fixes #1014 

## Proposed Changes

  - Avoid returning unnecessary data in the boards related requests
  - Add blur to the card parent element and remove it from the children
  - Disable drag and drop while the cards are hidden
  - Use floor instead of ceil when calculating the teams

This pull request closes #986 and closes #1024 and closes #1014

### DIFF
--- a/backend/src/modules/boards/interfaces/services/get.board.service.interface.ts
+++ b/backend/src/modules/boards/interfaces/services/get.board.service.interface.ts
@@ -21,7 +21,7 @@ export interface GetBoardServiceInterface {
 		size?: number
 	): Promise<BoardsAndPage | null>;
 
-	getBoardFromRepo(boardId: string): Promise<LeanDocument<BoardDocument> | null>;
+	getBoardFromRepo(boardId: string): Promise<Board | null>;
 
 	getBoardData(boardId: string): Promise<Board>;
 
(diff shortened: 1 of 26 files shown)


CatiaAntunes96 (approved):
LGTM 🚀

PR #1078, opened 2023-02-14
fix: add slack channel id to a team

### LINKED ISSUE #1073: [BUG]: keep the team even without a slack channel id
## Description

In the slack process, particularly in the teams build data, when we map the channels to add its "id" into the team we are excluding the team whose channel creation fails

### Expected result

What should happen?

In case a channel creation fails the team for that channel should not be excluded, this way this team will be added to the message posted in the master channel

mourabraz:
Fixes #1073

## Proposed Changes

  - keep the size of teams unchanged, even if a channel was not created in `SlackCommunicationApplication@createAllChannels`

@nunocaseiro 

Note: I still not using the factory to create fake data.

This pull request closes #1073


nunocaseiro (approved):
LGTM 🚀

PR #1091, opened 2023-02-15
fix: placeholder replaces card

### LINKED ISSUE #1090: [BUG]:  placeholder replaces card text
## Description

If a user tries to edit a card, the text in the textArea is being replaced with the placeholder. Also, when a responsible of a regular board with a team as participants, tries to change the column position, it occurs an error: error updating the board.

CatiaAntunes96:
Fixes #1090 

## Proposed Changes

  - When editing a card, the textarea value is the text of the card and not the placeholder (previous behavior)
  - On a regular board, when the "default card text" is active, the placeholder also acts as value for the textarea
  - Fix toggle when user toggles the default car text button
  - Add team to UpdateBoardType's schema and pass team id field when updating a column position

This pull request closes #1090


nunocaseiro (approved):
LGTM 🚀

Issue #742, opened 2022-12-23
[FEATURE]: Regular board - Public/Private


nunocaseiro:
## Summary

The sub-boards and the main board in a split retro should be always private unless the user is participating
In the regular board the board/team admin should be able to make the board public or private
If it is public, the system should handle the guests by creating a new instance in the boardUser
In the top right corner should be present a section with bubbles for the guests

## User story it relates to

#743

### Basic example

<img width="335" alt="image" src="https://user-images.githubusercontent.com/90208434/209343006-65cc9863-7474-4351-8b2c-14ab654e6d6e.png">

![image](https://user-images.githubusercontent.com/90208434/209342915-64baf85d-05a4-408c-bd58-782940ac202e.png)

PR #1093, opened 2023-02-16
feat: regular board public private

patricia-mdias:
Relates to #742 

## Proposed Changes
  - Create guest User and Board User endpoint, given a board.
  - Create only Board User endpoint, given an user and board.
  - Get public board endpoint.
  - Get public status endpoint, to check if board is public.


nunocaseiro (approved):

nunocaseiro (approved):
LGTM 🚀 GJ 🔥

PR #1166, opened 2023-02-28
Fix: regular retro column name

### LINKED ISSUE #1150: [BUG] Regular retro column name
## Description
As a user I should be able to edit column names

### Current behavior
- Editing the column name from the column itself has a 15 character limit, while in the settings menu it has a 30 character limit
- Characters that are longer than the standard ASCII characters will overflow the column
- Previous column name is being stored after changing it from a different place (changing it from the settings menu will not change it from the column menu and vice-versa)

https://user-images.githubusercontent.com/87712174/221160793-a1ddab15-1541-4587-938a-627dfff1d722.mov

CatiaAntunes96:
Fixes #1150 

## Proposed Changes

  - Fix maximum of column title chars to 15 letters
  - Fix input value when column name is changed (settings and column options)
  - Add overflow-text: ellipsis property for when column name is too long
  - Add tooltip to when is not possible to display all the column name
  - Add horizontal scroll on columns container for when it is not possible to display all of them on the screen 
  

This pull request closes #1150

### DIFF
--- a/frontend/src/components/Board/Column/Column.tsx
+++ b/frontend/src/components/Board/Column/Column.tsx
@@ -10,6 +10,7 @@ import { filteredColumnsState } from '@/store/board/atoms/filterColumns';
 import { countColumnCards } from '@/helper/board/countCards';
 import Icon from '@/components/Primitives/Icon';
 import Tooltip from '@/components/Primitives/Tooltip';
+import { useResizeDetector } from 'react-resize-detector';
 import AddCardOrComment from '../AddCardOrComment';
 import CardsList from './CardsList';
 import SortMenu from './partials/SortMenu';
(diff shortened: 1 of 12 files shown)


frontend/src/components/Board/Column/Column.tsx
```diff
@@ -130,13 +144,29 @@ const Column = React.memo<ColumMemoProps>(
...
+                        <TitleContainer
+                          css={{
+                            width: hasMoreThanThreeColumns ? '$130' : '$237',
+                          }}
+                        >
+                          {showTooltip ? (
+                            <Tooltip content={title}>
+                              <Title heading="4" ref={ref}>
+                                {title}
+                              </Title>
```
nunocaseiro, line 162:
```suggestion
                              <Title heading="4" ref={ref}>
                                {title}
                              </Title>
```

As a suggestion, you can extract this code into a new component that can be reusable when needed in this file

nunocaseiro (changes requested):

nunocaseiro (approved):

PR #1177, opened 2023-03-02
feat: get board guard

patricia-mdias:
## Proposed Changes

  - Create guard to avoid get board requests of a private board by user without access.


backend/src/libs/guards/getBoardPermissions.guard.ts
```diff
@@ -0,0 +1,49 @@
...
+		private readonly reflector: Reflector,
+		@Inject(Boards.TYPES.services.GetBoardService)
+		private getBoardService: GetBoardServiceInterface
+	) {}
+
+	async canActivate(context: ExecutionContext) {
+		const permissions = this.reflector.get<string>('permissions', context.getHandler());
+		const request = context.switchToHttp().getRequest();
+
+		const user = request.user;
```
nunocaseiro, line 27:
```suggestion
		const { isAnonymous, isSAdmin } = request.user;
```

backend/src/libs/guards/getBoardPermissions.guard.ts
```diff
@@ -0,0 +1,49 @@
...
+		try {
+			const { isPublic, team } = await this.getBoardService.getBoardData(boardId);
+			const boardUserFound = await this.getBoardService.getBoardUsers(boardId, user._id);
+
+			if (isPublic || boardUserFound.length) {
+				return true;
+			}
+
+			if (!boardUserFound) {
+				const teamUser = (team as Team).users.find(
```
nunocaseiro, line 39:
```suggestion
				const { role } = (team as Team).users.find(
```

nunocaseiro (changes requested):

nunocaseiro (approved):

Issue #1064, opened 2023-02-09
[FEATURE]: Add minimum time for the timer


geomarb:
## Summary

When setting the timer, it should have a minimum time (30 seconds).

## User story it relates to

As a user when I set the timer I don't want to be allowed to set it for less than 30 seconds.

### Basic example

When I set the timer now it allows me to set zero as the minimum time now.

### Motivation

Because there is no sense in setting a timer with less than 30 seconds by now. The timer was created to limit the time for specific tasks, like writing on the boards, voting, commenting, discussing actions, and so on. With that said it does not make sense allowing set the timer with less than a certain amount of time.

PR #1210, opened 2023-03-09
feat: add minimum time for the timer

geomarb:
Relates to:
- **#1064**

## Proposed Changes

This PR turns the timer (`useTimer`) more configurable. It is now possible to set up the following:

- **`START_MINUTES`**: (*default value: **`5`***) default minutes for all timers
- **`START_SECONDS`**: (*default value **`0`***) default minutes for all timers
- **`JUMP_SECONDS`**: (*default value: **`5`***) increase/decrease seconds when setting the duration
- **`JUMP_MINUTES`**: (*default value: **`1`***) increase/decrease minutes when setting the duration
- **`MIN_SECONDS`**: (*default value: **`30`***) minimum seconds accepted when setting the duration
- **`MIN_MINUTES`**: (*default value: **`0`***)  minimum minuted accepted when setting the duration
- **`MAX_SECONDS`**: (*default value: **`60 - JUMP_SECONDS`***)  maximum seconds to set, it is related to `JUMP_SECONDS` in this case it will `55` seconds, if `JUMP_SECONDS` is `10` it must be `50`and so on, otherwise it will be over `60` seconds.
- **`MAX_MINUTES`**: (*default value: **`59`***) maximum minutes allowed for the timer if in any case, it needs to be greater than `99` the timer's design and its CSS must be changed as well
 
**Note**: The `logger` was removed from the `events` in the backend, it actually should not be present in the `events` since if it is there it will be sent to the frontend inside the event. 

**Note 2**: The `previousStatus` was completely removed since wasn't necessary.

This pull request closes:
- **#1064**


nunocaseiro (approved):
LGTM 🚀

PR #1251, opened 2023-03-15
chore: update next and react

nunocaseiro:
## Proposed Changes

  - Update next and react


nunocaseiro:
@rpvsilva , can you check if the workflows are correctly modified?

rpvsilva:
> @rpvsilva , can you check if the workflows are correctly modified?

Yup, it's fine

Issue #1262, opened 2023-03-19
[BUG]: stop timer method is called unecessary 


nunocaseiro:
## Description

The stop timer method is called when the user opens a board page and it throws an exception

<img width="1257" alt="image" src="https://user-images.githubusercontent.com/90208434/226204540-cc5adce3-dbeb-45f6-ba0f-dc05ba635c85.png">

PR #1291, opened 2023-03-22
fix: 1262 bug stop timer method is called unnecessarily

geomarb:
Relates to:
- #1262 

Fixes:
- #1262 

## Proposed Changes

  - In the `useTimer` (frontend) the `isActiveTimer` was not checking correctly the function `isTimerPause` it was missing the parenthesis then it was always returning `true` causing the execution of `timerStop()` function. This was the main fix in this PR.
  - In the `backend` when stopping the timer it was not checking if the timer was found (if it exists) and if it was already stopped, now it is checking, and logging it as a `warn` when the timer was not found or was found and was already stopped, it won't throw an error, it will just log it as a `warn`.

Other changes:
  - The `default` was removed from some types, interfaces and enum
  - The `constants` and helper functions were moved out from `useTimer`
  - It is now treating when a user is trying to change the status of the timer for the status it already was in those cases it will log a `warn` (it won't throw an error or log as error since it shouldn't break the API)
  
This pull request closes #
- #1262


nunocaseiro (approved):
LGTM 🚀

PR #1306, opened 2023-03-24
refactor: create card service to usecase

### LINKED ISSUE #1294: [REFACTOR]: create.card.service to useCase and remove logic from card controller
## Summary

Currently, the Cards Controller contains some logic that should be moved to a UseCase. This will help to keep the code more organized and maintainable. Additionally, we need to refactor the create.card.service to start using useCase

### Proposed changes
- Move the card functionality logic from the Cards Controller to a UseCase.

GoncaloCanteiro:
Relates to:
- #1294

## Proposed Changes

  - Move controller logic to use case
  - Refactor service to use case
  - Remove create card service 
  

This pull request closes #1294


backend/src/modules/cards/dto/useCase/response/create-card-res.use-case.dto.ts
```diff
@@ -0,0 +1,13 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsNotEmpty } from 'class-validator';
+import Card from 'src/modules/cards/entities/card.schema';
+
+export default class CreateCardResUseCaseDto {
```
nunocaseiro, line 5:
```suggestion
export default class CardCreationPresenter {
```


Great job in doing this. The objects that are returned to the client from the controller are called "presenters" and the dto are the opposite. So I suggest adding a new folder in the card's folder called "presenters" and then move this class there

backend/src/modules/cards/dto/useCase/params/create-card.use-case.dto.ts
```diff
@@ -0,0 +1,19 @@
...
+
+	@ApiProperty()
+	@IsNotEmpty()
+	@IsString()
+	userId: string;
+
+	@ApiProperty()
+	@IsNotEmpty()
+	createCardDto: CreateCardDto;
+}
```
nunocaseiro, line 19: This just a dto, not a parameter validator. The params are the fields that are extracted from the endpoint path. So I think you don't need the `params` folder here

nunocaseiro (changes requested):

backend/src/modules/cards/dto/useCase/params/create-card.use-case.dto.ts
nunocaseiro: You can leave the useCase folder if it makes sense to you. I'm ok with that

backend/src/modules/cards/dto/useCase/response/create-card-res.use-case.dto.ts
nunocaseiro: Yes, I noticed the return type is from the use case to the controller, but I think it should also be a presenter

nunocaseiro (approved):
LGTM 🚀

PR #1318, opened 2023-03-29
refactor: useTeams custom hook

### LINKED ISSUE #1295: [BUG]: React-Query Improvements
## Description

 - Improve React-Query with some functionalities provided by the library that aren't implemented in our project.
 - Those functionalities are:
   - Stable Query Client
   - Refetch on Mount
   - Stale Time for API Requests
 
The idea here is to make a Prefetch on the ServerSide and have the information available on the Frontend when the Page Loads.
This should only make 1 request to the backend.

However when a change occurs, we should update the Query to reflect that change, instead of refetching the information again.

StereoPT:
Relates to:
- #1295 

Fixes:
- #1295 

## Proposed Changes

  - Refactored useTeams Custom Hook
    - Split this hook into singular hooks to improve usage
  - Removed duplication of server state by removing server state from recoil and only using react-query
  - Improved requests by adding optimistic updates on some queries.

This pull request closes #1295


frontend/src/hooks/teams/useUpdateTeamUser.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { toastState } from '@/store/toast/atom/toast.atom';
+
+import { TEAMS_KEY } from '.';
+
+const useUpdateTeamUser = (teamId: string, userId?: string) => {
+  const queryClient = useQueryClient();
+  const setToastState = useSetRecoilState(toastState);
+
+  return useMutation(updateTeamUser, {
+    onSuccess: async (res: TeamUserUpdate) => {
```
nunocaseiro, line 18: I think this operation can be done `onMutate` method because you have all the needed information before the request but if you disagree we can follow this approach as the request is fast to be handled

nunocaseiro (changes requested):

nunocaseiro (approved):
LGTM 🚀

JoaoSaIvador (approved):
LGTM ⚡

Issue #1264, opened 2023-03-19
[BUG]: Create split board after team creation with less than 4 members


nunocaseiro:
## Description

If the user creates a team with just one member then he tries to create a split board by clicking on "create the first team board" on the team's page. The error message is shown as it is supposed to say the team doesn't have a sufficient number of members. After this, if the user adds at least 3 members to the team and if he tries to create a split board with the same procedure the error message doesn't disappear until the user refreshes the browser. The error message shouldn't be shown after the team has sufficient members 

Also, the color of the plus button's is blue instead of black

<img width="1262" alt="image" src="https://user-images.githubusercontent.com/90208434/226210720-3bdf10c0-8e68-4d98-9f86-6ddd5a39eea0.png">

<img width="1053" alt="image" src="https://user-images.githubusercontent.com/90208434/226210873-f332e0db-542f-4a1b-8a31-84a921e4360a.png">

PR #1327, opened 2023-03-31
refactor: remove unnecessary teams recoil state

### LINKED ISSUE #1326: [REFACTOR]: remove unnecessary teams recoil state
## Description

- Remove the unnecessary teams recoil state that could be switched to react query instead.
- This affects the state that is mainly being used just to read data from and not being manipulated. 
- Instead of having 2 places with the same data to read from, it's better to just have the data in the cache and read it using react query.

JoaoSaIvador:
Relates to:
- #1326 
- #1264 

Fixes:
- #1326
- #1264 

## Proposed Changes

  - Remove the unnecessary teams recoil state that could be switched to react query instead.
  - This affects the state that is mainly being used just to read data from and not being manipulated.
  - Instead of having 2 places with the same data to read from, it's better to just have the data in the cache and read it using react query.

This pull request closes #1326 and #1264


StereoPT (approved):
LGTM  🗝️

PR #1335, opened 2023-04-03
refactor: pass href to styled components anchor

juniorboos:
Fixes:
image

## Proposed Changes
  - Add `passRef` and `legacyBehaviour` as recommended by [Next.js docs](https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag).


dsousa12 (approved):

juniorboos:
> This was introduced when we upgraded to Next 13. There might be more instances of this type of problem

Yeah, this was all I could find as a direct anchor child. Other places where nesting `button` and other components, but I couldn't find any warning on those.

CatiaBarroco-xgeeks (approved):
🎸 LGTM

PR #1360, opened 2023-04-05
fix: board not found redirect

### LINKED ISSUE #1317: [BUG]: accessing a non existing board
## Description

When an unauthenticated user tries to access a non-existing board via URL, the app should redirect to the landing page.
If the user is authenticated the app should stay on the same page.

Instead of that, a 400 error is shown. It happens because the public request fails when the `boardId` isn't a `mongoId` string. The endpoint method on the backend has a BaseParam that checks if the `boardId` has a `mongoId` type and as it isn't true, a `400` error is thrown.

This can be solved on the frontend.

- In the frontend, the call of the request on the server side can be wrapped by a try-catch and the app should redirect to the landing page or to the 404 page based on the user's session.

### Steps to reproduce

- Accessing the `http://localhost/boards/123`

### Actual result

image

juniorboos:
Fixes:
- #1317

## Proposed Changes
  - Wrap `getPublicStatusRequest` inside the existing try-catch, as it will redirect the user to the proper page based on the session.

This pull request closes #1317

### DIFF
--- a/frontend/src/pages/boards/[boardId]/index.tsx
+++ b/frontend/src/pages/boards/[boardId]/index.tsx
@@ -44,11 +44,6 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
 
   const session = await getSession(context);
 
-  if (boardId.includes('.map'))
-    return {
-      props: {},
-    };
-
   try {
     const boardIsPublic = await getPublicStatusRequest(boardId, context);
 
--- a/frontend/src/pages/boards/[boardId]/index.tsx
+++ b/frontend/src/pages/boards/[boardId]/index.tsx
@@ -49,24 +49,27 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
       props: {},
     };
 
-  const boardIsPublic = await getPublicStatusRequest(boardId, context);
-
-  // if board is public and user has no session
-  if (boardIsPublic && !session) {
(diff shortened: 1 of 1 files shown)


StereoPT:
LGTM 🗝️ 

However I think you can also remove the if condition that checks that the boardId includes a .map

juniorboos:
> LGTM 🗝️
> 
> However I think you can also remove the if condition that checks that the boardId includes a .map

Sure! Can you give me more context on why this was needed?

StereoPT:
> Sure! Can you give me more context on why this was needed?

From what I know, that was a one time fix for a problem that can't be reproduced.
However you can't find any mentions of a .map on a Mongo ObjectId. Therefore that condition is never going to be true.
@nunocaseiro can give you more information about why it was added. But I think it can be safely removed.

nunocaseiro:
> > Sure! Can you give me more context on why this was needed?
> 
> From what I know, that was a one time fix for a problem that can't be reproduced. However you can't find any mentions of a .map on a Mongo ObjectId. Therefore that condition is never going to be true. @nunocaseiro can give you more information about why it was added. But I think it can be safely removed.
> 
> Also, I think it's better to Redirect to the Boards Page instead of to the Dashboard

Yes, you can remove it. Somehow in a previous version, when refreshing the board page it triggered an error and that condition suppresses it

Issue #971, opened 2023-01-30
[FEATURE]: Responsive web design


nunocaseiro:
## Summary

Add some responsiveness to the website. It doesn't have to be mobile friendly, but it can have some behaviours that don't break the design as it currently does

PR #1388, opened 2023-04-13
feat: add Tiles responsiveness

juniorboos:
Relates to:
- #971

| **Old**                                                                                                                                    | **New**                                                                                                                                    |
|--------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------|
| image  | image  |
| image  | image  |
| image | image |

## Proposed Changes
  - Move `Tile` styles to a variant instead of a const.
  - Fix text underneath `TileBlob` with `relative` position.
  - Move `padding` to `StyledTile` so it can be used by the `TileArrow` as well.
  - Hide overflow so that the `TileBlob` svg fits inside the container.
  - Replace unnecessary `absolute` `TileArrow` positioning with `alignSelf`.
  - Add `textShadow` so that the text can be visible when above `TileBlob`.
  - Replace `grid` display with `flex` for Tiles to be responsive according to the container's width.


frontend/src/components/Dashboard/Tiles/styles.tsx
```diff
@@ -1,9 +1,11 @@
...
 import { styled } from '@/styles/stitches/stitches.config';
 
-const TileContainer = styled('section', {
-  display: 'grid',
+const TileContainer = styled(Flex, {
   gap: '$29',
-  gridTemplateColumns: 'repeat(3, 1fr)',
+  flexWrap: 'wrap',
+
+  '> *': { flexGrow: 1 },
```
StereoPT, line 8: Can you explain the reason why use `flexGrow` instead of just ´flex: 1`?

Also, what are your opinions on having a fixed `minWidth`?

juniorboos: Yep, that can also work! I'm not sure why `minWidth` would be useful as I'm expecting `flex` to decide whether it should wrap or not based on the available space. I'm also enforcing that the text shouldn't break, so it will always have a readable width.

frontend/src/components/Dashboard/Tiles/styles.tsx
```diff
@@ -1,9 +1,11 @@
+import Flex from '@/components/Primitives/Layout/Flex/Flex';
 import { styled } from '@/styles/stitches/stitches.config';
 
-const TileContainer = styled('section', {
-  display: 'grid',
+const TileContainer = styled(Flex, {
```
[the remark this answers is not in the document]
juniorboos: Fair enough! I changed in 40d24ac.

frontend/src/components/Dashboard/Tiles/styles.tsx
StereoPT: I was thinking that so the Tiles would always have the same width.

juniorboos: Isn't that what `flex: 1` is doing? Not sure I understood.

nunocaseiro (approved):
LGMT 🚀

frontend/src/components/Dashboard/Tiles/Tile/styles.tsx
```diff
@@ -1,34 +1,61 @@
...
+
+  width: '100px !important',
+  height: '100px !important',
+  top: '0',
+
+  variants: {
+    color: {
+      yellow: {
+        width: '127px !important',
+        height: '76px !important',
```
nunocaseiro, line 46: is it possible to use the stitches tokens instead of `px`?

nunocaseiro (changes requested):

frontend/src/components/Dashboard/Tiles/styles.tsx
StereoPT: The idea of the minWidth is to guarantee that each tile has exactly the same size.
As you can see, the `Your teams` is smaller than the `Active Members`

image

It was just a design idea

juniorboos: Ooh, I see what you mean! I think it's hard to define a `minWidth` here as the `width` is defined by the text's length. So even if we define it as `300px` for example, there could be a Tile bigger than that because of the text inside it, and then they wouldn't be the same size once again. The only way to enforce the same size in this situations is using `maxWidth`, but I don't think that's scalable.

StereoPT (approved):
LGTM  🗝️

tutods (approved):

nunocaseiro (approved):

PR #1402, opened 2023-04-14
fix: accessing deleted board

### LINKED ISSUE #1382: [BUG] Accessing deleted board
## Description
As a user I should not be able to access boards that no longer exist

### Current behavior
If I'm inside a board while it gets deleted, network is spammed with 403 errors

Image

### Proposed change
Boot the user out of the board if this board gets deleted

StereoPT:
Relates to:
- #1382 

Fixes:
- #1382 

## Screen Recording

https://user-images.githubusercontent.com/8330038/232097658-251614c0-de7a-4c19-a4e9-9dffca954e7b.mov

## Proposed Changes

  - When a user is in a board that is deleted:
    - He receives a message via `socket`
    - He is redirected to the `board-deleted` page 
    - This will happen in the Main Board and also on the Sub-Boards

This pull request closes #1382


backend/src/modules/boards/controller/boards.controller.ts
```diff
@@ -331,10 +332,17 @@ export default class BoardsController {
...
+
+		const result = await this.deleteBoardUseCase.execute({ boardId, completionHandler });
 
 		if (socketId && teamId) {
+			// This socket emit doesn't have a recieving function on the Frontend.
+			// Is it really necessary?
 			this.socketService.sendUpdatedBoards(socketId, teamId);
+
 			this.socketService.sendUpdatedBoard(boardId, socketId);
 		}
```
nunocaseiro, line 347: Yes, this can be deleted

nunocaseiro:
I'm not sure if is this the best solution for this feature. We can simply redirect the user to the board's page if the getBoard request receives an error

nunocaseiro (changes requested):

PR #1405, opened 2023-04-17
feat: add Button responsiveness

juniorboos:
Relates to:
- #971

| Old                                                                                                                 | New                                                                                                                 |
|---------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------|
|  |  |
|   |   |

## Proposed Changes

  - Removed unnecessary styles
     - Replaced multiple svg `height` attributes with a single `aspectRatio`.
     - Removed `:hover` selector when it comes to `cursor`.
  - Fixed `Link`'s `height` being bigger than the `Button` inside it when the flex container wraps.
  - Removed fixed `height` so that the Button can properly wrap (`height` should be dynamic, controlled by paddings and content).
     - Added a transparent border so that all variants have the same height (updated paddings accordingly).
  - Added `minWidth` to svg to fix a bug where the icons were shrinking in smaller containers.
  - Added `xl` size variant to Storybook.


frontend/src/components/Primitives/Inputs/Button/Button.tsx
```diff
@@ -210,96 +200,66 @@ const Button = styled('button', {
...
-        '& svg': {
-          height: '$32',
-          width: '$32',
-        },
+        py: '$14',
+        '& svg': { width: '$32', minWidth: '$32' },
         '& span': {
           height: '$32',
           width: '$32',
         },
```
juniorboos, line 212: I couldn't find any occurrences of a `span` inside a `Button`, so I guess this could be removed too. I need more context on it tho.

StereoPT: A `span` inside a button can be a `Text` Primitive.

frontend/src/components/Primitives/Inputs/Button/Button.tsx
```diff
@@ -8,24 +8,17 @@ const Button = styled('button', {
   display: 'flex',
   justifyContent: 'center',
   alignItems: 'center',
-  border: 'none',
+  border: '2px solid transparent',
```
StereoPT, line 11: I like this approach. Good Job.

StereoPT (approved):
LGTM  🗝️

In the Button Primitive, I've added a comment at the Top that might be an interesting idea.
Divide the Buttons into 2 different styled components:
 - Have a Base Button that has all the Main Styles
 - And then have both a Button and also an IconButton
 
This would allow us to have buttons that only have an icon and control their size independently.

frontend/src/components/Primitives/Inputs/Button/Button.tsx
juniorboos: Oh right! I'll keep it then.

PR #1415, opened 2023-04-19
fix: fix responsible field on divided boards

### LINKED ISSUE #1381: [BUG] Regular board no responsible
## Description
As a user I should be able to create and delete regular boards for me and my team members

### Current behavior
As a super admin, it's possible to directly remove myself from the board, leaving no responsible on the board, this board will only be visible to the participants but no one can delete it
As a regular user, I can select another member to join my board, when I remove them, the frontend is actually removing me, making it possible to create regular boards for other members without board responsible, this board is impossible to delete.

### Steps to reproduce

https://user-images.githubusercontent.com/87712174/231185928-e4f9c5f6-4b05-4ad5-a3d9-6344b0993d6e.mov

### Proposed changes
Perhaps the backend should check these things, instead of the frontend?
Make sure for regular boards to always have at least 1 responsible, if there are more than 1 make sure the board creator is also included and is a board responsible
Super admins should be able to see these boards and be able to delete them

CatiaAntunes96:
Fixes:
- #1381 

## Proposed Changes

  - Fixed bug related to responsible field being empty on subBoards

This pull request closes #1381


frontend/src/pages/boards/newSplitBoard.tsx
```diff
@@ -185,6 +190,8 @@ const NewSplitBoard: NextPage = () => {
       role: boardUser.role,
     }));
 
+    console.log(newDividedBoards);
+
```
nunocaseiro, line 194:
```suggestion
```

nunocaseiro (changes requested):

nunocaseiro (approved):

Issue #1365, opened 2023-04-05
[REFACTOR]:  change module types for constants


CatiaAntunes96:
## Description

Change module types for constants and import only the necessary constants for each file.

### Motivation

Apply good practices regarding file naming and organisation.

PR #1426, opened 2023-04-24
refactor: change votes module types for constants

CatiaAntunes96:
Relates to:
- #1365 

## Proposed Changes

  - Change types for constants on votes module


patricia-mdias (approved):
LGTM! 🚀

PR #1432, opened 2023-04-24
refactor: extract usecases of columns module

### LINKED ISSUE #1427: [REFACTOR]: extract usecases of columns module
## Description

For the columns module, we should extract all the use cases from the services and update the tests related to them.

CatiaAntunes96:
Relates to:
- #1427 

## Proposed Changes

  - Extract useCases from updateColumn service
  - Refactor tests to use the library @golevelup/ts-jest

This pull request closes #1427


backend/src/modules/columns/applications/delete-cards-from-column.use-case.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+
+		getBoardServiceMock.getBoardById.mockResolvedValue(board);
+	});
+
+	it('should be defined', () => {
+		expect(useCase).toBeDefined();
+	});
+
+	describe('delete cards from column', () => {
+		it('should return a updated board without cards on the column', async () => {
```
patricia-mdias, line 74:
```suggestion
		it('should return an updated board without cards on the column', async () => {
```

patricia-mdias (approved):
LGTM! 🚀

PR #1434, opened 2023-04-26
refactor: change users module types for constants

CatiaAntunes96:
Relates to:
- #1365 

## Proposed Changes

  - Change types for constants on the users module


backend/src/modules/auth/applications/register-guest-user.use-case.ts
```diff
@@ -10,7 +10,7 @@ import { InsertFailedException } from 'src/libs/exceptions/insertFailedBadReques
 @Injectable()
 export default class RegisterGuestUserUseCase implements RegisterGuestUserUseCaseInterface {
 	constructor(
-		@Inject(TYPES.services.CreateUserService)
+		@Inject(CREATE_USER_SERVICE)
 		private createUserService: CreateUserServiceInterface,
```
GoncaloCanteiro, line 14: Since you are changing these files you can add `readonly` here and in the other files

GoncaloCanteiro (approved):
LGTM! 🚀

PR #1445, opened 2023-04-28
refactor: change types for constants on azure module

CatiaAntunes96:
Relates to:
- #1365 

## Proposed Changes

  - Change types for constants on azure module


nunocaseiro (approved):
LGTM 🚀

PR #1446, opened 2023-04-28
feat: add users responsiveness

juniorboos:
Relates to:
- #971

| Old                                                                                                             | New                                                                                                             |
|-----------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|
| image | image |
| image | image |
| image | image |

## Proposed Changes
  - Refactored `Badge` to be a `span` instead of a `div`, as it makes more sense for an inline component.
  - Refactored `UserItem` to be responsive.
  - Fixed `Button` with `isIcon` in a `disabled` state.
  - Fixed `ConfigurationSwitch` `disabled` state.
  - Added `width` prop to `UncontrolledInput`.
  - Removed unnecessary wrapping `div` in `UserItemActions`.
  - Replaced `margins` with `gap`.
  - Replaced `455` size token with `456` to follow the 8px rule.


nunocaseiro (approved):
LGTM 🚀

PR #1459, opened 2023-05-03
fix: query function without return

juniorboos:
image

## Proposed Changes
  - Remove braces to return the function result.


JoaoSaIvador (approved):
LGTM ⚡

PR #1583, opened 2024-11-13
feat: issue 945 soft delete users

### LINKED ISSUE #945: [FEATURE]: Soft delete users
## Summary

The delete user feature should set the field  "isDeleted" to true instead of removing the user from the database. However, the user must be removed from the teams in which he is participating. (the field is already present in the user's schema)
All the database queries that request users should consider this field. So those queries shouldn't fetch the users that have the `isDeleted == true`. 
When a deleted user tries to login or register he shouldn't complete those operations with success. In the future, we can also add a feature to the super admin "revive" or force delete the user if it makes sense

### Motivation

To maintain the board users available we shouldn't remove the user reference by deleting the user in the database

joaofrparreira:
Relates to:
- #945 

## Proposed Changes

  - Add Soft-delete mongoose plugin (can be used for other schemas)
  - Add soft-delete plugin to User Schema
  - Remove user from not submitted boards only, to maintain history
  - All the query's consider `isDeleted` field to not present data when it's `true`, unless the query specifies their presentation
  - Delete-user-case now also deletes the user from the open boards
  

@nunocaseiro 

This pull request closes #945


nunocaseiro (approved):
LGTM 🚀

PR #1595, opened 2024-11-26
fix: issue-1593- validate access token on login register

### LINKED ISSUE #1593: [BUG]: Validate AccessToken on login/register
## Description

AD access token is not being validated, and it should be to prevent the use of forged tokens

### Steps to reproduce

Try to login with a forged access token.

### Expected result

User should not be able to login or register with an invalid access token

### Actual result

User might login with a forged access token.

joaofrparreira:
Fixes:
- #1593 

## Proposed Changes

  - Validate AD access token on user login/register

This pull request closes #1593

### DIFF
--- a/backend/src/modules/azure/applications/register-or-login.azure.use-case.ts
+++ b/backend/src/modules/azure/applications/register-or-login.azure.use-case.ts
@@ -100,7 +100,7 @@ export class RegisterOrLoginAzureUseCase implements RegisterOrLoginAzureUseCaseI
 			userToAuthenticate.avatar = avatarUrl;
 		}
 
-		return await signIn(userToAuthenticate, this.getTokenService, 'azure');
+		return signIn(userToAuthenticate, this.getTokenService, 'azure');
 	}
 
 	private async getUserPhoto(user: User) {
--- a/backend/src/modules/azure/applications/register-or-login.azure.use-case.ts
+++ b/backend/src/modules/azure/applications/register-or-login.azure.use-case.ts
@@ -1,9 +1,8 @@
-import { Inject, Injectable } from '@nestjs/common';
+import { Inject, Injectable, Logger } from '@nestjs/common';
 import { RegisterOrLoginAzureUseCaseInterface } from '../interfaces/applications/register-or-login.azure.use-case.interface';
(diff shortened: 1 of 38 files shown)


backend/src/modules/azure/applications/register-or-login.azure.use-case.ts
```diff
@@ -75,7 +100,7 @@ export class RegisterOrLoginAzureUseCase implements RegisterOrLoginAzureUseCaseI
 			userToAuthenticate.avatar = avatarUrl;
 		}
 
-		return signIn(userToAuthenticate, this.getTokenService, 'azure');
+		return await signIn(userToAuthenticate, this.getTokenService, 'azure');
```
nunocaseiro, line 103:
```suggestion
		return signIn(userToAuthenticate, this.getTokenService, 'azure');
```
I think you don't need the `await` here

backend/src/modules/azure/services/auth.azure.service.ts
```diff
@@ -1,16 +1,19 @@
...
 
 export type AzureUserFound = {
 	id: string;
 	mail: string;
 	displayName: string;
 	userPrincipalName: string;
 	createdDateTime: Date;
+	accountEnabled: boolean;
+	deletedDateTime: Date | null;
 };
```
nunocaseiro, line 17: I know this was already here but what do you think about moving this to a "types" folder? Does it make sense to you?

backend/src/modules/azure/services/auth.azure.service.ts
```diff
@@ -68,4 +79,37 @@ export default class AuthAzureService implements AuthAzureServiceInterface {
 	fetchUserPhoto(userId: string) {
 		return this.graphClient.api(`/users/${userId}/photo/$value`).get();
 	}
+
+	async getADUsers(): Promise<Array<AzureUserDTO>> {
```
nunocaseiro, line 83: Is this method being used or are you just preparing the code to the next feature?

nunocaseiro (changes requested):

backend/src/modules/azure/services/auth.azure.service.ts
joaofrparreira: makes all the sense.

PR #1602, opened 2024-12-04
feat:  automatically add users to team during synchronization

### LINKED ISSUE #1601: [FEATURE]: Auto-add users to pre-defined team
## Summary

Automatically add users to pre-defined team (if set in env vars) during AD-App users synchronization

joaofrparreira:
Relates to:
- #1601 

## Proposed Changes

  - Automatically add's new users to a team defined in env vars during AD-App users synchronization cron job
  - Changes insertion of users from each one to a user batch

This pull request closes #1601

### DIFF
--- a/backend/src/modules/azure/schedules/synchronize-ad-users.cron.azure.use-case.ts
+++ b/backend/src/modules/azure/schedules/synchronize-ad-users.cron.azure.use-case.ts
@@ -46,6 +46,7 @@ export class SynchronizeADUsersCronUseCase implements SynchronizeADUsersCronUseC
 		try {
 			let team;
 			const teamName = this.configService.get('AD_SYNCHRONIZATION_AUTO_ADD_USER_TEAM_NAME');
+			const userEmailDomain = this.configService.get('AD_SYNCHRONIZATION_EMAIL_DOMAIN');
 
 			if (teamName !== '') {
 				team = await this.getTeamByNameUseCase.execute(teamName);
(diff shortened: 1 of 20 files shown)


backend/src/modules/azure/schedules/synchronize-ad-users.cron.azure.use-case.ts
```diff
@@ -43,6 +66,8 @@ export class SynchronizeADUsersCronUseCase implements SynchronizeADUsersCronUseC
 				throw new Error('Split app users list is empty.');
 			}
 
+			const usersAppFiltered = usersApp.filter((user) => user.email.endsWith('@xgeeks.com'));
```
nunocaseiro, line 69: What do you think if we add a new env var for the team email's domain? e.g. "@xgeeks.com"

You've added an env var to define the team name so maybe it also makes sense to add the email domain

backend/src/modules/azure/schedules/synchronize-ad-users.cron.azure.use-case.ts
```diff
@@ -91,28 +116,51 @@ export class SynchronizeADUsersCronUseCase implements SynchronizeADUsersCronUseC
...
-			} catch (err) {
-				this.logger.error(
-					`An error as occurred while creating user with email '${user.mail}' through the syncronize AD Users Cron. Message: ${err.message}`
-				);
+				};
+			});
+
+			const createdUsers = await this.createUserService.createMany(usersToCreate);
+
+			if (typeof team !== 'undefined') {
```
nunocaseiro, line 145:
```suggestion
			if (team) {
```

Do we really need to check the type of the team?

nunocaseiro (changes requested):

backend/src/modules/azure/schedules/synchronize-ad-users.cron.azure.use-case.ts
joaofrparreira: It makes all the sense

backend/src/modules/azure/schedules/synchronize-ad-users.cron.azure.use-case.ts
joaofrparreira: Force of habit, in this case bot are equally fine to use.

nunocaseiro (approved):
LGTM 🚀

